PR #2, opened 2022-08-11
Use and_then

Masynchin:
Hello, @trimental!

I just took a look at source code and found this little opportunity for improvement. That is it.


trimental:
Thanks for the first PR!

PR #4, opened 2022-08-12
Have line-delta scrolling shift by 30 pixels per line

### LINKED ISSUE #3: Crashes on scroll wheel
```
thread 'main' panicked at 'not implemented', src/main.rs:172:30
```

(Linux, Wayland)

LovecraftianHorror:
Partially resolves #3

This could be improved by calculating a more appropriate value to scroll by line, but this change at least makes things more usable when line-delta scrolling is used

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -150,44 +150,28 @@ impl Inlyne {
                 }
                 Event::WindowEvent { event, .. } => match event {
                     WindowEvent::CloseRequested => *control_flow = ControlFlow::Exit,
-                    WindowEvent::MouseWheel { delta, .. } => match delta {
-                        MouseScrollDelta::PixelDelta(pos) => {
-                            {
-                                let screen_height = self.renderer.screen_height();
-                                if self.renderer.reserved_height > screen_height {
-                                    self.renderer.scroll_y -= pos.y as f32;
+                    WindowEvent::MouseWheel { delta, .. } => {
+                        let y_pixel_shift = match delta {
+                            MouseScrollDelta::PixelDelta(pos) => pos.y as f32,
+                            // Arbitrarily pick x30 as the number of pixels to shift per line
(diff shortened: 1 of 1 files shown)


trimental:
@LovecraftianHorror Considering the default text size is 16. which is scaled by `hidpi_scale` (which is around 2. on macOS) Could you test using `16. * self.renderer.hidpi_scale` for different scale factors using `inlyne --scale`.

LovecraftianHorror:
Here's a demo when using `* 16.0 * self.renderer.hidpi_scale` on my system so you can get a better idea of how it looks. This is with `--scale 1.5` on X.org which has notoriously annoying DPI issues, so it's possible that wayland handles this better

https://user-images.githubusercontent.com/30302768/184280131-06bca6be-2664-44d4-83f9-7f55b262913d.mp4

I tried to showcase how much a "single line" is in the beginning which is why it's so jerky

trimental:
It's hard to get a sense of how scrolling feels through a video, although it seems to scroll by about one default sized line.
Did `16.0 * self.renderer.hidpi_scale` feel good at different scales?

As long as you can move around fast enough with it, I think that approach would work better.

LovecraftianHorror:
The scrolling feels a bit sluggish compared to other programs on my computer. The last half of the clip I sent is me scrolling pretty aggressively to give it some more context

Switching to `32.0 * self.renderer.hidpi_scale` seems to better match how much other programs on my computer scroll

trimental:
@LovecraftianHorror Well we can always update it in the future but for now I think `32 * hidpi_scale` is a good default.

Thank you for the PR! Especially since testing on x11 is a pain on my MacBook.

lnicola:
I think it's fine. I think wheel scroll defaults to three lines on Windows. I get about five in Firefox and 2.25 in VS Code :shrug:.

PR #5, opened 2022-08-12
Improve error message when file path isnt a file

LovecraftianHorror:
This change just tries to provide a friendlier message when there is a failure reading from the provided file

_Before_

```console
$ inlyne missing_file.md
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }', src/main.rs:890:60
```

_After_

```console
$ inlyne missing_file.md
Error: Error reading file at "missing_file.md"

Caused by:
    No such file or directory (os error 2)
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -69,6 +69,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "anyhow"
+version = "1.0.61"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


trimental:
@LovecraftianHorror Definitely looks a lot clearer and less of an information overload than just an unwrap. Thanks 👍. 

One tiny thing to change is that instead of stating error twice, could the message be changed to `Error: Could not read file at "{location}"`

LovecraftianHorror:
Done and done!

PR #6, opened 2022-08-12
Remove `debug = true` for release builds

LovecraftianHorror:
It looks like this was added when when doing some profiling. Not including debug info makes the binary size a lot smaller for Linux users


trimental:
Thanks for the PR!

Yep definitely forgot to take this out after debugging.

PR #9, opened 2022-08-13
Initial codeblock syntax highlighting support

### LINKED ISSUE #8: Support codeblock highlighting
It would be really nice to have syntax highlighting support for `inlyne`. I haven't delved into the html rendering side, but getting `comrak` to support adding colored spans in codeblocks is as simple as

```patch
diff --git a/src/main.rs b/src/main.rs
index c14b060..7677e81 100644
--- a/src/main.rs
+++ b/src/main.rs
@@ -928,7 +928,12 @@ fn main() {
         options.extension.strikethrough = true;
         options.parse.smart = true;
         options.render.unsafe_ = true;
-        let htmlified = markdown_to_html(&md_string, &options);
+
+        let mut plugins = comrak::ComrakPlugins::default();
+        let adapter = comrak::plugins::syntect::SyntectAdapter::new("base16-ocean.dark");
+        plugins.render.codefence_syntax_highlighter = Some(&adapter);
+
+        let htmlified = comrak::markdown_to_html_with_plugins(&md_string, &options, &plugins);
 
         input.push_back(
             Tendril::from_str(&htmlified)
```

The theme is configurable (passed to `SyntextAdapter`), and this renders html that looks like so when viewed in my browser

image

and here is the raw HTML for the first block

```html
<pre style="background-color:#2b303b;"><code class="language-js">
<span style="color:#65737e;">// cat.js
</span><span style="color:#b48ead;">import </span><span style="color:#c0c5ce;">{ </span><span style="color:#bf616a;">resolve </span><span style="color:#c0c5ce;">} </span><span style="color:#b48ead;">from </span><span style="color:#c0c5ce;">&quot;</span><span style="color:#a3be8c;">path</span><span style="color:#c0c5ce;">&quot;;
</span><span style="color:#b48ead;">import </span><span style="color:#c0c5ce;">{ </span><span style="color:#bf616a;">write</span><span style="color:#c0c5ce;">, </span><span style="color:#bf616a;">stdout</span><span style="color:#c0c5ce;">, </span><span style="color:#bf616a;">file</span><span style="color:#c0c5ce;">, </span><span style="color:#bf616a;">argv </span><span style="color:#c0c5ce;">} </span><span style="color:#b48ead;">from </span><span style="color:#c0c5ce;">&quot;</span><span style="color:#a3be8c;">bun</span><span style="color:#c0c5ce;">&quot;;
</span><span style="color:#c0c5ce;">
</span><span style="color:#b48ead;">const </span><span style="color:#c0c5ce;">path = </span><span style="color:#8fa1b3;">resolve</span><span style="color:#c0c5ce;">(</span><span style="color:#bf616a;">argv</span><span style="color:#c0c5ce;">.</span><span style="color:#8fa1b3;">at</span><span style="color:#c0c5ce;">(-</span><span style="color:#d08770;">1</span><span style="color:#c0c5ce;">));
</span><span style="color:#c0c5ce;">
</span><span style="color:#b48ead;">await </span><span style="color:#8fa1b3;">write</span><span style="color:#c0c5ce;">(
</span><span style="color:#c0c5ce;">  </span><span style="color:#65737e;">// stdout is a Blob
</span><span style="color:#c0c5ce;">  </span><span style="color:#bf616a;">stdout</span><span style="color:#c0c5ce;">,
</span><span style="color:#c0c5ce;">  </span><span style="color:#65737e;">// file(path) returns a Blob - https://developer.mozilla.org/en-US/docs/Web/API/Blob
</span><span style="color:#c0c5ce;">  </span><span style="color:#8fa1b3;">file</span><span style="color:#c0c5ce;">(path)
</span><span style="color:#c0c5ce;">);
</span><span style="color:#c0c5ce;">
</span><span style="color:#65737e;">// bun ./cat.js ./path-to-file
</span>
</code></pre>
```

I would be happy to do the work rendering the colored spans if you think this change would be worthwhile

LovecraftianHorror:
Resolves #8

_The `comrak` feature tweaking is to get rid of the default `clap` feature which is only useful for the `comrak` binary_

This implements the initial support for syntax highlighting code blocks, but it seems there are some bugs in how some spans with just whitespace are handled (e.g. the lines starting with `async` and `return`) which I would appreciate some help troubleshooting since the `TokenPrinter` is a big ball of mutable state that is pretty hard to follow.

Here is some of the syntax highlighted code where the indentation is having some obvious issues. The indentation issues are reproducible when just having `comrak` include colored spans in codeblocks without any of the code for respecting the span's color

_Before_

image

_After_

image

---

Also worth noting that the HTML portion of the code snippet should also be indented, but isn't in either this PR or `main`

```js
// requires Bun v0.1.0 or later
// react-ssr.tsx
import { renderToReadableStream } from "react-dom/server";

const dt = new Intl.DateTimeFormat();

export default {
  port: 3000,
  async fetch(request: Request) {
    return new Response(
      await renderToReadableStream(
        <html>
          <head>
            <title>Hello World</title>
          </head>
          <body>
            <h1>Hello from React!</h1>
            <p>The date is {dt.format(new Date())}</p>
          </body>
        </html>
      )
    );
  },
};

// bun react-ssr.tsx
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -60,15 +60,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ansi_term"
-version = "0.12.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d52a9bb7ec0cf484c551830a7ce27bd20d67eac647e1befb56b0be4ee39a55d2"
-dependencies = [
- "winapi",
-]
-
-[[package]]
 name = "anyhow"
 version = "1.0.61"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -258,21 +249,6 @@ checksum = "fff857943da45f546682664a79488be82e69e43c1a7a2307679ab9afb3a66d2e"
 
 [[package]]
 name = "clap"
-version = "2.34.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a0610544180c38b88101fecf2dd634b174a62eef6946f84dfc6a7127512b381c"
-dependencies = [
- "ansi_term",
- "atty",
- "bitflags",
- "strsim 0.8.0",
- "textwrap 0.11.0",
- "unicode-width",
- "vec_map",
-]
-
-[[package]]
-name = "clap"
 version = "3.2.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a3dbbb6653e7c55cc8595ad3e1f7be8f32aba4eb7ff7f0fd1163d4f3d137c0a9"
@@ -283,9 +259,9 @@ dependencies = [
  "clap_lex",
  "indexmap",
  "once_cell",
- "strsim 0.10.0",
+ "strsim",
  "termcolor",
- "textwrap 0.15.0",
+ "textwrap",
 ]
 
 [[package]]
@@ -373,7 +349,6 @@ version = "0.14.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "15bf1e432b302dc6236dd0db580d182ce520bb24af82d6462e2d7a5e0a31c50d"
 dependencies = [
- "clap 2.34.0",
  "entities",
  "lazy_static",
  "memchr",
@@ -599,7 +574,7 @@ dependencies = [
  "ident_case",
  "proc-macro2",
  "quote",
- "strsim 0.10.0",
+ "strsim",
  "syn",
 ]
 
@@ -1100,7 +1075,7 @@ version = "0.1.5"
 dependencies = [
  "anyhow",
  "bytemuck",
- "clap 3.2.16",
+ "clap",
  "comrak",
  "copypasta",
  "html5ever",
@@ -2544,12 +2519,6 @@ dependencies = [
 
 [[package]]
 name = "strsim"
-version = "0.8.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8ea5119cdb4c55b55d432abb513a0429384878c15dde60cc77b1c99de1a95a6a"
-
-[[package]]
-name = "strsim"
 version = "0.10.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "73473c0e59e6d5812c5dfe2a064a6444949f089e20eec9a2e5506596494e4623"
@@ -2639,15 +2608,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bab24d30b911b2376f3a13cc2cd443142f0c81dda04c118693e35b3835757755"
 dependencies = [
  "winapi-util",
-]
-
-[[package]]
-name = "textwrap"
-version = "0.11.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d326610f408c7a4eb6f51c37c330e496b08506c9457c9d34287ecc38809fb060"
-dependencies = [
- "unicode-width",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


trimental:
I tried in c746cbdc3ab9d7d82c51b1dcf8d335cfac41717a to clean up the html interpreter a bit as it is definitely a mess 😅. To briefly explain it now uses a `State` struct to track things like span color, text formatting options, links, etc. For some things I've stored this as an integer instead of boolean to track levels of depth, ie

```
<a href="cats.html">
<a href="dogs.html">
<p> a link </p> // Depth 2
</a>
<p> still a link </p> // Depth 1
</a>
// Depth 0
```

While element_stack stores elements that their children need access to, so <table> is stored on this stack so <td> <th> <tr> can access it.

About your the whitespace not being accounted for, its because wgpu_glyph only reports whitespaces between text and not at the start or ends so that the size is representative of text and not whitespace.

I've managed to get this
Screenshot 2022-08-15 at 12 46 18 am
when adding 
```
                if str == "\n" {
                    if self.state.text_options.pre_formatted >= 1 {
                        if !self.current_textbox.texts.is_empty() {
                            self.push_element(self.current_textbox.clone().into());
                            self.current_textbox.texts.clear();
                        }
                    } else if !self.current_textbox.texts.is_empty() {
                        self.current_textbox.texts.push(Text::new(
                            " ".to_string(),
                            self.hidpi_scale,
                            self.theme.text_color,
                        ));
                    }
                } else {
```
 to https://github.com/trimental/inlyne/commit/c746cbdc3ab9d7d82c51b1dcf8d335cfac41717a#diff-60405aaba3dc351a041fce3e90f17b63dee9a214bc9c35ae6a5126d003100e55R448

trimental:
By the way code colouring makes the application feel a lot more complete, once again thank you for taking the time to help out with this.

src/main.rs
```diff
@@ -686,6 +688,27 @@ impl TokenSink for TokenPrinter {
...
+                                    if let Some(hex_str) = styles
+                                        .split(';')
+                                        .find_map(|style| style.strip_prefix("color:#"))
+                                    {
+                                        if let Ok(hex) = u32::from_str_radix(hex_str, 16) {
+                                            let [_, r, g, b] = hex.to_be_bytes();
+                                            let r = r as f32 / 255.0;
+                                            let g = g as f32 / 255.0;
+                                            let b = b as f32 / 255.0;
+                                            self.span_color = [r, g, b, 1.0];
```
trimental, line 706: Using the function `hex_to_linear_rgba` from colour.rs like `self.state.span_color = hex_to_linear_rgba(hex);` helps to convert the srgb web colour into rgba linear which wgpu uses. We need to do this conversion to stop colors washing out and appearing muted.

src/main.rs
```diff
@@ -916,7 +938,12 @@ fn main() -> anyhow::Result<()> {
         options.extension.strikethrough = true;
         options.parse.smart = true;
         options.render.unsafe_ = true;
-        let htmlified = markdown_to_html(&md_string, &options);
+
+        let mut plugins = comrak::ComrakPlugins::default();
+        let adapter = comrak::plugins::syntect::SyntectAdapter::new("InspiredGitHub");
```
trimental, line 943: It would also be nice to be able to change this depending on the theme. It should be as simple as adding `pub code_highlighter: &'static str` to `Theme` and replacing `InspiredGithub` with `theme.code_highlighter`.

I think `InspiredGithub` looks really good for light theme, and I'm a fan of `base16-mocha.dark`. Please let me know if you have any other suggestions.

LovecraftianHorror:
Thanks for the explanation on how the html interpreter works, it really helped clear things up a lot!

If you already have a branch that has the codeblocks highlighting changes on top of the html interpreter refactor then feel free to push it here and I'll address the other review comments!

src/main.rs
LovecraftianHorror: That sounds good to me!

Eventually we can make it configurable so that users can pick any of the themes that `syntect` supports. `comrak` doesn't expose a way to set a theme other than the ones included by default AFAICT, but that still give users a handful of different options at least

trimental:
@LovecraftianHorror I've just forced my branch of your PR for you to take a look at on your device. 

I also added a padding function to `TextBox` that just adds height regardless of whether the text is whitespace. By adding a textbox with a padding of 18., it's the same as adding an empty line. This fixes the fact that the code boxes used to ignore empty lines.

LovecraftianHorror:
Just pulled it down, and it looks great!

Thanks for helping out so much with this PR. It looks like that resolves both of the review comments

trimental:
@LovecraftianHorror Team work makes the dream work. I'll spend tomorrow looking for bugs and hopefully release the next version tomorrow night 👍

LovecraftianHorror:
Sounds great!

Most of my contributing will be bursts on weekends, so I won't try pushing through more changes before the next release :smile:

PR #10, opened 2022-08-14
Add a config file for persistent settings

LovecraftianHorror:
This change adds in the ability to use a config file located at `<config_dir>/inlyne/inlyne.toml` for setting persistent settings. The current supported settings are just to set different defaults for the `--scale` and `--theme` instead of only being able to pass them on the command line. Having a config file for just two settings seems overkill, but it's reasonable that there will be more settings added over time (custom color themes, different syntax highlighting, custom keybinds, etc.)

This did involve refactoring the CLI arg parsing for two reasons

1. Distinguishing between an unset `--theme` and `--theme light`
2. Changing the default values listed in help messages based on settings in the `inlyne.toml`

Managing both of those involved switching to the builder API, so that we have some lower level control

---

Here's a peek at how the help messages look

_Before_

```text
inlyne 0.1.5
trimental
Introducing Inlyne, a GPU powered yet browsless tool to help you quickly view markdown files in the
blink of an eye.

USAGE:
    inlyne [OPTIONS] <FILE_PATH>

ARGS:
    <FILE_PATH>

OPTIONS:
    -h, --help             Print help information
    -s, --scale <SCALE>
    -t, --theme <THEME>    [default: light] [possible values: dark, light]
    -V, --version          Print version information
```

_After no config_

```text
inlyne 0.1.5
trimental
Introducing Inlyne, a GPU powered yet browsless tool to help you quickly view markdown files in the
blink of an eye.

USAGE:
    inlyne [OPTIONS] <FILE>

ARGS:
    <FILE>    Path to the markdown file

OPTIONS:
    -h, --help             Print help information
    -s, --scale <scale>    Factor to scale rendered file by [default: Window's scale factor]
    -t, --theme <theme>    Theme to use when rendering [default: light] [possible values: dark,
                           light]
    -V, --version          Print version information
```

_After with config_

```toml
scale = 1.5
theme = "Dark"
```

```text
trimental
Introducing Inlyne, a GPU powered yet browsless tool to help you quickly view markdown files in the
blink of an eye.

USAGE:
    inlyne [OPTIONS] <FILE>

ARGS:
    <FILE>    Path to the markdown file

OPTIONS:
    -h, --help             Print help information
    -s, --scale <scale>    Factor to scale rendered file by [default: 1.5]
    -t, --theme <theme>    Theme to use when rendering [default: dark] [possible values: dark,
                           light]
    -V, --version          Print version information
```


src/cli.rs
```diff
@@ -0,0 +1,236 @@
...
+
+    fn gen_args(args: Vec<&str>) -> Vec<OsString> {
+        std::iter::once("inlyne")
+            .chain(args.into_iter())
+            .map(OsString::from)
+            .collect()
+    }
+
+    #[test]
+    fn debug_assert() {
```
trimental, line 157: What is the purpose of the same test twice with a different scaling help text. Especially since "1.5" or "window's scale factor" isn't what the program would use at runtime for the help text.

trimental:
Great improvement. Having a configuration file is definitely a must have and saves having to add arguments to the end of every command. 

I only have that one comment above, everything else seems to be in good shape 👍

src/cli.rs
LovecraftianHorror: It doesn't really make sense to have that test twice, I agree

I switched it to just one test using the same help message that would be passed in (even though the specific contents won't have an impact)

LovecraftianHorror:
Addressed the review comment and fixed merge conflicts!

trimental:
@LovecraftianHorror excellent, thank you very much! Once https://github.com/trimental/inlyne/pull/9 lands, I'll make the next release. I prefer quick releases when it comes to binaries, and it'll give you a chance to get the tag based AUR package out.

PR #11, opened 2022-08-14
Swap `reqwest` out for `ureq`

LovecraftianHorror:
`reqwest` has a pretty heavy dependency tree considering it is only used once in `inlyne`. On top of that it still pulls in `tokio` when  the rest of the application isn't `async`

Swapping to `ureq` shaves off a healthy amount of dependencies which reduces build times and reduces the final binary size by ~1 MiB for me


trimental:
I also agree that a blocking request crate fits the purposes of this crate more than an async based one. Thank you for suggesting this 👍.

trimental:
I'd also like to add that images load faster for me, at least on debug when using this PR. A nice bonus.

PR #12, opened 2022-08-15
Switch image min and mipmap filters to linear

LovecraftianHorror:
I know I said I was done with changes before the next release, but I was trying to figure out why some of the images had some obvious fuzziness. Setting an anisotropy clamp cleared it up, and from what I can tell this seems to have minimal impact on GPU usage (not surprising since this isn't an image heavy program)

All of these images are from using a window that is one quarter the size of my monitor with a scale of 1.5 since this exacerbates the issue. The second image is portion from the last image in bun's README

_Before_

image

image

_After_

image

image

### DIFF
--- a/src/image.rs
+++ b/src/image.rs
@@ -371,6 +371,7 @@ impl ImageRenderer {
             mag_filter: wgpu::FilterMode::Linear,
             min_filter: wgpu::FilterMode::Nearest,
             mipmap_filter: wgpu::FilterMode::Nearest,
+            anisotropy_clamp: Some(std::num::NonZeroU8::new(4).expect("4 is not 0")),
             ..Default::default()
         });
         Self {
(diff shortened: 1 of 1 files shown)


trimental:
@LovecraftianHorror Could you try instead switching 

```
            min_filter: wgpu::FilterMode::Nearest,
            mipmap_filter: wgpu::FilterMode::Nearest,
```
to Linear and testing out whether it fixes the issue for you as well. I think that anisotropic filtering is only really important when viewing things at a weird non perpendicular perspective, like in video games. The issue I think is going on here is that its trying to sample whole pixels instead of blending them when downscaling.

LovecraftianHorror:
Thanks for the tip (I'm not super familiar with GPU related things)! That looks like it also clears up the image quality. I'll go ahead and push that change instead

trimental:
Perfect, good spot on that. The images look a lot better now 👍.

PR #15, opened 2022-08-16
Allow for specifying alignment in `<div>`s

LovecraftianHorror:
This PR was motivated by the decoration at the top of [`insta`'s README](https://github.com/mitsuhiko/insta/blob/master/README.md)

```html
<div align="center">
 <img src="https://github.com/mitsuhiko/insta/blob/master/assets/logo.png?raw=true" width="250" height="250">
 <p><strong>insta: a snapshot testing library for Rust</strong></p>
</div>
```

The two aspects of this PR are

1. Pushing div alignment onto the element stack
2. Only set `current_textbox` alignment when found

_Before_

image

_After_

image


trimental:
Looks good to me. Thanks for this 👍.

PR #18, opened 2022-08-16
Tweak inline code colors

LovecraftianHorror:
The inline code color looks a bit out of place now that there code blocks have syntax highlighting

This PR changes the inline code colors based by picking an arbitrary color from the syntax highlighting. I just tried to pick a color that was both distinct from regular text color, link color, and something that would look like a clicked link color (I know the color doesn't change when they're clicked)

Feel free to bikeshed the colors!

_Before Light_

image

_After Light_

image

_Before Dark_

image

_After Dark_

image

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -30,7 +30,7 @@ pub const DARK_DEFAULT: Theme = Theme {
         b: 0.0086,
         a: 1.0,
     },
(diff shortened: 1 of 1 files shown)


trimental:
Eventually I'd like to allow customising themes in the configuration file but this is much less of an attack on the senses for the default configuration, thanks.

PR #20, opened 2022-08-17
Support custom themes in `inlyne.toml`

LovecraftianHorror:
> Eventually I'd like to allow customising themes in the configuration file

I agree! How about now for eventually?

Here's how the dark theme looks in `inlyne.toml.sample`

image


trimental:
Thanks great work as always @LovecraftianHorror 💯. Your code is always a lot clearer than mine which makes reviewing it a breeze. 

The only thing I would change is `clear-color` to `background-color` or maybe `background-canvas-color` if you think that's too general. I think `clear-color` made sense when it was hardcoded as it described where in wgpu it was used, but most users probably don't need to know this.

LovecraftianHorror:
> Thanks great work as always @LovecraftianHorror 💯. Your code is always a lot clearer than mine which makes reviewing it a breeze. 

Thanks! I think you're code is totally fine considering I've been able to make all these changes though ♥️

> The only thing I would change is `clear-color` to `background-color` or maybe `background-canvas-color`

`background-color` sounds good! I was confused by what `clear-color` meant until I was working in these changes, so user's probably would be too

Feel free to push the changes to this branch if you want. I won't be able to get to it for ~10 more hours

PR #22, opened 2022-08-17
Add font configuration

trimental:
Adds font configuration through the `[font-options]` section of the configuration file.

```
# # Change the appearance of text with font options
# [font-options]
# # Name of font for regular text
# regular-font = "Chalkduster"
# # Name of font for monospace text such as code
# monospace-font = "Monaco"
```

Screenshot 2022-08-17 at 11 45 17 pm

Failing to find the requested font it gives an error similar to 
```
Error: No font found for monospace font

Caused by:
    no font found
```

However in the case no font is provided it falls back to the system default monospace and sans serif font.


trimental:
@LovecraftianHorror It would be nice if you could take a quick once over to make sure I haven't done anything in `src/opts` that's a worse way then you intended. Any other comments/suggestions are also welcome :smile:

LovecraftianHorror:
Really great to see this!

I'll give it a review after I get of work!

trimental:
Going to experiment with using a different library as on my flamegraph font-kit's `select_best_match` uses 25% of cpu power and delays startup considerably. Hopefully font-kit just implements this shockingly and other libraries do it better.

LovecraftianHorror:
No idea if we can use it, but `usvg` uses `fontdb` which has [a method for loading system fonts](https://docs.rs/fontdb/latest/fontdb/struct.Database.html#method.load_system_fonts)

trimental:
@LovecraftianHorror For some reason that performance penalty is only when using `PingFang SC`. Theres almost no penalty for other fonts. 

I did see that but it seems very hacky and probably doesn't support bolding or italics, I remember doing something similar when I made https://github.com/Smithay/andrew for smithay-client-toolkit and trying to find font directories yourself isn't the best idea. I'm happy to just use font kit for now considering its just one chinese font, but I'll keep it in mind if any future issues come up.

src/fonts.rs
```diff
@@ -0,0 +1,102 @@
...
+
+impl fmt::Display for FontError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match &self {
+            FontError::CopyingFontData => write!(f, "Error copying font data"),
+        }
+    }
+}
+
+impl Error for FontError {}
```
LovecraftianHorror, line 26:
```suggestion
```

It looks like all of this is dead code, so it can just be removed

LovecraftianHorror (approved):
These changes look great! I only had one small nit

> For some reason that performance penalty is only when using PingFang SC. Theres almost no penalty for other fonts.

Yeah I'm noticing it with my `ttf-iosevka-nerd` font as well. When running with defaults it's really fast though. My best guess is that whatever matching logic they use slows down a lot when it encounters a font with a lot of glyphs

When I set my monospace font to Iosevka it looks like each call to `.select_best_match()` takes a bit over 200ms (edit: each call to `.select_best_match()` for the monospace family). From a quick peek browse of `font-kit` I think I see a way that things can be sped up, but I won't have time to try hacking things together till this weekend. Since this is just a perf issue and it only happens when specifying some specific fonts I don't see any reason why it should block merging this PR :+1:

---

_Also this change drops the binary size by almost a meg because it's no longer bundling the fonts :partying_face:_

PR #23, opened 2022-08-19
Fixup link clicking

LovecraftianHorror:
The main motivation behind all of this is that clicking links without moving the mouse wouldn't actually open the link. This is because link clicking was implemented in a position change event, so pressing and releasing the LMB without moving the cursor wouldn't trigger it

This was fixed by moving the link clicking logic into the LMB press event. Most of the work here was refactoring to decouple the hover element checking from link clicking to avoid duplicating a bunch of logic

The scrollbar could use similar changes since it suffers from the same problem of clicks not registering unless the cursor is moved while clicking, but this is all I have time for tonight

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -18,7 +18,6 @@ use crate::text::Text;
 use color::Theme;
 use renderer::Positioned;
 use renderer::{Renderer, Spacer};
-use utils::HoverInfo;
 use utils::Rect;
 
 use anyhow::Context;
@@ -116,10 +115,10 @@ impl Inlyne {
     }
 
     pub fn run(mut self) {
-        let mut click_scheduled = false;
         let mut scrollbar_held = false;
         let mut mouse_down = false;
         let mut modifiers = ModifiersState::empty();
+        let mut last_loc = (0.0, 0.0);
         self.event_loop.run(move |event, _, control_flow| {
             *control_flow = ControlFlow::Wait;
             match event {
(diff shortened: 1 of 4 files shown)


trimental:
Thanks for this and your time taken. I've taken a quick look this morning and everything looks pretty good to me! I'll do a proper review tonight. 

There are a few small changes I have about redundant redraws and when the selection is set but if you're happy I can push these to the PR to save you time.

src/main.rs
```diff
@@ -280,35 +243,56 @@ impl Inlyne {
...
+                                    Hoverable::Text(Text { link, .. }) => link,
+                                };
+
+                                if let Some(link) = maybe_link {
+                                    if open::that(link).is_err() {
+                                        if let Some(anchor_pos) =
+                                            self.renderer.positioner.anchors.get(link)
+                                        {
+                                            self.renderer.set_scroll_y(*anchor_pos);
+                                            self.window.request_redraw();
```
trimental, line 281: This is the redundant redraw I mentioned. Redraws are by far the most expensive operation and I don't think this is needed.

src/main.rs
```diff
@@ -280,35 +243,56 @@ impl Inlyne {
                             if !scrollbar_held {
                                 scrollbar_held = true;
                             }
-                        } else if click_scheduled && !jumped {
+                        } else if self.renderer.selection.is_none() {
```
trimental, line 246: Although updating tuple values is very cheap, I think it makes more sense and is more efficient if selection is only set on a click pressed event. To prevent the issue of selection when jumping 

```
+     } else if self.renderer.selection.is_none() {
+         // Only set selection when not over link
+         self.renderer.selection = Some((last_loc, last_loc));
        }
+ } else if self.renderer.selection.is_none() {
+     self.renderer.selection = Some((last_loc, last_loc));
+ }
``` 
can be added [here](https://github.com/trimental/inlyne/pull/23/files#diff-42cb6807ad74b3e201c5a7ca98b911c5fa08380e942be6e4ac5807f8377f87fcR284)

LovecraftianHorror:
> but if you're happy I can push these to the PR to save you time

That's fine by me!

PR #26, opened 2022-08-20
Give me window resizes or give me death

### LINKED ISSUE #25: Bad window resize performance with large READMEs on some systems
For some reason on my system resizing the window has really _really_ bad performance

## Demo

https://user-images.githubusercontent.com/30302768/185724443-d6ed6c2c-0760-48cf-ad69-a46e35a53f8d.mp4

As you can see it processes 100+ events for `WindowEvent::Resized` that all take ~20ms along with one beefy event for `Event::RedrawRequested` that takes ~16s

If I do small shifts to the window then it's not a problem. It's only when a lot of `WindowEvent::Resized` get queued up that `Event::RedrawRequested` takes a _long_ time

_What's the CPU usage look like?_

It takes up all of one logical core for the whole duration

_Memory usage?_

Peak usage is usually over 10 GiB :smiling_face_with_tear: 

This does not go down after the resize finishes

_What if you do it twice?_

The `WindowEvent::Resized`s consistently each takes ~20ms while the `Event::RedrawRequested` takes around 1.5 seconds

_Is this a recent problem?_

Nope. I can reproduce it on commits back before I even started contributing

_Does this happen with smaller READMEs?_

No! The issue seems to only exist when the README is big enough that another `WindowEvent::Resized` gets queued up before the last one finishes

## Profiles

_Both profiles are from the same situation shown in the video aka one resize and then killing the program_

### CPU usage

image

As you can see the vast majority of the time is spent in glyph cache related things

### Memory usage

image

_Note: 12.3GB peak memory usage_

Unsurprisingly also dominated by glyph cache related things

## Investigation

The vast majority of the time for the `WindowEvent::Resized` related calls are taken up from this `.glyph_bounds()` call

https://github.com/trimental/inlyne/blob/14422b0d3a2a7893ee483df8b5c6ae9fc239d4e9/src/text.rs#L100

The vast majority of the time for the `Event::RedrawRequested` call is taken up by this `.draw_queued_with_transform()` call

https://github.com/trimental/inlyne/blob/14422b0d3a2a7893ee483df8b5c6ae9fc239d4e9/src/renderer.rs#L572-L597

Two things both reduce the time taken for `Event::RedrawRequested` to ~1.5s and reduced the peak memory usage from ~12GB to ~4.5GB

1. Adding [`.draw_cache_position_tolerance(1.0)`](https://docs.rs/wgpu_glyph/latest/wgpu_glyph/struct.GlyphBrushBuilder.html#method.draw_cache_position_tolerance) to the glyph brush builder (This ignores sub-pixel position changes when caching glyphs)
2. Setting the `scale` to 1.0 (My hunch here was that the `hidpi_scale` was inconsistently used somewhere causing duplicate glyphs to be added to the cache)

And sadly I haven't found anything that makes `WindowEvent::Resized` process faster. I think filtering consecutive queued `WindowEvent::Resized` events could help since there is no reason to process a resize if there is another resize already queued after it. I did a hack to only process every 10th `WindowEvent::Resized` which made things refresh smoothly aside from having obvious graphical issues. Changing the event loop to allow inspecting queued up events would be very non-trivial though. My best idea on how to handle it would be to send events over a channel to another thread that actually processes them, but this involves a lot of changes from lifetimes, callbacks, etc.

All of these seem like they're trying to work around some underlying issue that I can't find though

## Pending questions

Does something queue up each time there is a `WindowEvent::Resized`? It seems that way since the time taken up by `Event::RedrawRequested` depends on the number of `WindowEvent::Resized` that have run before it

What's causing the glyph cache to blow up so much? Even with the hack above 4.5 GB peak memory usage still seems very high

LovecraftianHorror:
Resolves #25 

The comment linked below covers things pretty well. The gist is that we can lazily handle window resizing since we will get a `RedrawRequested` afterwards. This prevents a bunch of `WindowResized`s from clogging up the event queue

https://github.com/trimental/inlyne/compare/main...LovecraftianHorror:give-me-window-resizes-or-give-me-death?expand=1#diff-42cb6807ad74b3e201c5a7ca98b911c5fa08380e942be6e4ac5807f8377f87fcR191-R216

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -219,8 +219,6 @@ impl Inlyne {
                 None => return,
             };
 
-            println!("{event:?}");
-
             match event {
                 RelevantEvent::UserEvent(inlyne_event) => match inlyne_event {
                     InlyneEvent::Reposition => {
--- a/src/main.rs
+++ b/src/main.rs
@@ -59,6 +59,7 @@ enum RelevantEvent {
     LmbReleased,
     ModifiersChanged(ModifiersState),
     KeyPressed(VirtualKeyCode),
+    MainEventsCleared,
 }
 
 impl RelevantEvent {
@@ -94,6 +95,7 @@ impl RelevantEvent {
                 } => Some(Self::KeyPressed(keycode)),
                 _ => None,
             },
+            Event::MainEventsCleared => Some(Self::MainEventsCleared),
             _ => None,
         }
     }
@@ -174,11 +176,39 @@ impl Inlyne {
     }
 
     pub fn run(mut self) {
+        // Why do we handle resizes lazily like this?
+        //
+        // For a bit of background. `winit`'s event loop has separate tiers of events. It won't move
+        // on to the next tier until all of the events from the current tier have been processed and
+        // (here's the kicker) events can keep coming in while all of this is happening. Here are
+        // the tiers for reference
+        //
+        // 1. Window events, User events, Device events
+        // --- `MainEventsCleared` sent ---
+        // 2. Redraw windows (goes back to 1. after this)
+        //
+        // The obvious issue here is that expensive events from one tier continually coming in will
+        // block us from moving on to the next tier
+        //
+        // How does this matter in practice?
+        //
+        // Dragging to resize a window is represented as a mix of window events (`WindowResized` and
+        // `CursorMoved`) followed by a final `MainEventsCleared`. With large READMEs a window
+        // resize will be expensive because it has to reposition many elements. This means that
+        // dragging a window will queue up a lot of window resizes before we can even redraw a
+        // window, but it's pointless to calculate a window resize if there is already another size
+        // pending
+        //
+        // Instead we lazily store the size and only reposition elements and request a redraw when
+        // we recieve a `MainEventsCleared` indicating that we've finished recieveing the first tier
+        // of events. This prevents us from clogging up the queue with a bunch of costly resizes.
+        // For more information take a look at
+        // https://github.com/trimental/inlyne/issues/25
+        let mut pending_resize = None;
         let mut scrollbar_held = false;
         let mut mouse_down = false;
         let mut modifiers = ModifiersState::empty();
         let mut last_loc = (0.0, 0.0);
-        let mut pending_resize = None;
         let event_loop_proxy = self.event_loop.create_proxy();
         self.event_loop.run(move |winit_event, _, control_flow| {
             *control_flow = ControlFlow::Wait;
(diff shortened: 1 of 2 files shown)


LovecraftianHorror:
Oh also

# _weeeeeeee_

https://user-images.githubusercontent.com/30302768/185731052-86bd26a4-ec8c-41e5-b705-70ad6bb4f52b.mp4

trimental:
Unfortunately this doesn't work in its current state on macOS since theres no other events to process after a window resize. The result is that the window stays unrefreshed and resized until you do another winit action like click the window. 

I've also experimented with a seperate rendering thread to help this issue but the one thing that always got me was resizing! The thing is that theres never going to be a `ResizedFinished` event since that would require psychic abilities to predict, so I think we have to treat every resize as if it's our last. Although I wonder what event on X11 is causing you to resize and whether its making it instantly resize after every resize event.

LovecraftianHorror:
Bleh, looks like we're running into platform specific differences since it appears that `RedrawRequested` gets called automatically for me (unless something else triggers it that I'm not aware of). Here is a snippet of me printing _all_ `ReleventEvent`s when dragging a window to resize

```text
CursorMoved(PhysicalPosition { x: 615.0, y: 475.0 })
WindowResized(PhysicalSize { width: 615, height: 477 })
WindowResized(PhysicalSize { width: 615, height: 475 })
WindowResized(PhysicalSize { width: 615, height: 473 })
CursorMoved(PhysicalPosition { x: 614.0, y: 469.0 })
WindowResized(PhysicalSize { width: 614, height: 471 })
WindowResized(PhysicalSize { width: 614, height: 469 })
WindowResized(PhysicalSize { width: 614, height: 468 })
CursorMoved(PhysicalPosition { x: 613.0, y: 463.0 })
WindowResized(PhysicalSize { width: 613, height: 465 })
RedrawRequested
WindowResized(PhysicalSize { width: 613, height: 464 })
WindowResized(PhysicalSize { width: 613, height: 462 })
WindowResized(PhysicalSize { width: 613, height: 460 })
WindowResized(PhysicalSize { width: 613, height: 459 })
WindowResized(PhysicalSize { width: 613, height: 457 })
WindowResized(PhysicalSize { width: 613, height: 455 })
WindowResized(PhysicalSize { width: 614, height: 454 })
RedrawRequested
WindowResized(PhysicalSize { width: 614, height: 452 })
WindowResized(PhysicalSize { width: 614, height: 451 })
WindowResized(PhysicalSize { width: 614, height: 450 })
WindowResized(PhysicalSize { width: 615, height: 449 })
WindowResized(PhysicalSize { width: 616, height: 448 })
WindowResized(PhysicalSize { width: 616, height: 447 })
WindowResized(PhysicalSize { width: 617, height: 446 })
RedrawRequested
WindowResized(PhysicalSize { width: 618, height: 445 })
WindowResized(PhysicalSize { width: 619, height: 444 })
WindowResized(PhysicalSize { width: 620, height: 444 })
WindowResized(PhysicalSize { width: 621, height: 443 })
WindowResized(PhysicalSize { width: 623, height: 443 })
WindowResized(PhysicalSize { width: 624, height: 443 })
RedrawRequested
```

When dragging the window to shrink it will pick up on cursor movements on the window. When expanding it will just be `WindowResized` and `RedrawRequested`

I've just pushed changes to trigger the resize off of a `MainEventsCleared` event which is supposed to fire between finishing all the window/user/device events and a redraw requested event. Let me know if this causes the window to stay refreshed for you

trimental:
Mmm it does work however I don't see any difference in what its doing then before. I thought that `request_redraw` already queues redraws to occur once the main events are cleared?

LovecraftianHorror:
I had an issue with endlessly looping redraws before, but i just realized that i was requesting a redraw on every redraw event 🤦‍♂️

I'll switch things around to simplify them after i wake up

LovecraftianHorror:
Actually I think main events cleared is still the ideal place to handle this since it doesn't involve having to special case any future window/user/device events that we may handle

Can you clarify what you mean by "before"?

LovecraftianHorror:
I think all the issues with this PR are resolved at this point

trimental:
With my previous comment I meant to say that on macOS it showed that resizes were being triggered immediately after pending_resize was set. So

```
[src/main.rs:174] "pending resize" = "pending resize"
[src/main.rs:392] "actual resize" = "actual resize"
[src/main.rs:174] "pending resize" = "pending resize"
[src/main.rs:392] "actual resize" = "actual resize"
```
but after looking closer I've seen a few 
```
[src/main.rs:174] "pending resize" = "pending resize"
[src/main.rs:174] "pending resize" = "pending resize"
[src/main.rs:392] "actual resize" = "actual resize"
```
so this PR is working 🎊. The only thing I'm unsure about is whether we still need `RelevantEvent`, it seems to now just be a clone of `winit::Event` with less information, unless I'm missing something.

LovecraftianHorror:
Yeah `RelevantEvent` isn't very useful at this point. I'll go ahead and switch to using `Event` again

trimental:
Oh one last thing (sorry to be a pain) but there needs to be a `self.renderer.set_scroll_y(self.renderer.scroll_y);` in the resize handler because if you scroll to the bottom and then resize, everything moves up and your left outside of visible content bounds. `set_scroll_y` will make scroll_y be in bounds.

LovecraftianHorror:
Nice catch on that! Should be all good now :crossed_fingers:

trimental:
Looks good 👍

PR #30, opened 2022-08-21
Add filepath to title

LovecraftianHorror:
This changes the title to include the filepath to the file being viewed

I figured what people would want 90% of the time would be the relative path from the repo root, so that is what this PR implements. If it doesn't detect a repo root then it falls back to the full filepath

image

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -37,7 +37,7 @@ use winit::{
 };
 
 use std::collections::VecDeque;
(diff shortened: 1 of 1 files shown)


trimental:
Looks great! Finding the git and mercurial is a simple feature but makes the application feel intelligent. Thanks for all these PRs.

PR #31, opened 2022-08-21
Add initial logging support

LovecraftianHorror:
I opted for `env_logger` since I like the filtering options it provides, but I would be open to alternatives (I also like `tracing-subscriber`)

The env var for filtering is `INLYNE_LOG`. By default dependencies only show `Error`s while `inlyne` shows `Info` and up to keep the default output from being very noisy


trimental:
Definitely will be good in the future for debugging and knowing what's going on internally!

PR #32, opened 2022-08-21
Avoid copying image data when possible

LovecraftianHorror:
Tiny little change

Some of the raw image data can be pretty big (the largest image in bun's README is 55 MiB raw). This tiny change avoids copying the image data if the `DynamicImage` is already in the rgba8 format. This lowers the peak memory usage of loading bun's README by 55 MiB


trimental:
Good spot, didn't notice that `to_rgba8` always creates a copy. Thanks for this, memory usage is starting to look better ⚡

PR #33, opened 2022-08-23
Allow for customizing lines scrolled

### LINKED ISSUE #13: Scroll wheel too slow
Scrolling does not scroll in a direction as much as expected.

Platform: Windows

LovecraftianHorror:
Resolves #13 

This PR adjusts the default number of lines to scroll by to 3, and allows for setting a custom value with the `lines-to-scroll` field in `inlyne.toml`

### DIFF
--- a/inlyne.toml.sample
+++ b/inlyne.toml.sample
@@ -7,6 +7,11 @@
 # # Possible values: ["Light", "Dark"]
 # theme = "Dark"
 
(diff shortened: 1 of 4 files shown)


trimental:
Looks good to me. I think we can close https://github.com/trimental/inlyne/issues/13 now since it's up to the user, but 3 seems like a good default from testing using a mouse 👍

Issue #21, opened 2022-08-17
Support more language types and error display


Blue-Savage:
https://i.imgur.com/3w5h93E.png

There are two problems with the window in the image.
1, can't display text other than English
2, there is a chance that only part of the content is displayed. Both windows load the same file.

trimental:
With regards to the Chinese characters not being displayed, it's as simple as the bundled San Francisco font not supporting these characters. Bundling fonts definitely makes cross-platform easier but allowing the user to select their own font is something I'd like to support. 

If you could maybe link the file that causes this that would be a great help 😃

Blue-Savage:
You can try it yourselves. The display problem is easily reproducible.
https://github.com/Blue-Savage/temp/raw/main/inlyne.7z

trimental:
Strangely enough everything seems to be rendering for me, perhaps the Chinese is just blinding me to it though. I've included a screenshot of Chinese character rendering using PR https://github.com/trimental/inlyne/pull/22 to select `PingFang SC` as the default font.

<img width="1440" alt="Screenshot 2022-08-17 at 11 58 26 pm" src="https://user-images.githubusercontent.com/40879396/185187385-8de9308d-5d25-4e17-9740-d85e9b5af6b6.png">

trimental:
I'm closing this as resolved since font loading is up to the user and I can't replicate the display issue on the file provided. Please feel free to send any images of the rendering bug to reopen this!

PR #37, opened 2022-08-24
Live reloading (exciting)

trimental:
Definitely an exciting PR that adds live reloading by watching the file directory for changes and then sending new string data to the interpreter. Live reloading is one of the things I most wanted for inlyne so I'm glad to at least get a proof of concept PR out.

There a few graphical glitches I'm experiencing on macOS with partial rendering which could be the same as stated in https://github.com/trimental/inlyne/issues/21. Honestly after looking into it I'm tempted to think it might be a wgpu issue, as it seems it takes two redraws for things drawn previously to show to the screen. Maybe a swap chain issue? I'll have to look into it more.

@LovecraftianHorror If you have the time to quickly help test this on linux when you get a chance that would be a great help even if it's on the weekend ❤️ (I should hurry up and get my dev setup on asahi linux)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -109,7 +109,7 @@ checksum = "d9b39be18770d11421cdb1b9947a45dd3f37e93092cbf377614828a319d5fee8"
 dependencies = [
  "hermit-abi",
  "libc",
- "winapi",
+ "winapi 0.3.9",
 ]
 
 [[package]]
@@ -279,7 +279,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9fdf5e01086b6be750428ba4a40619f847eb2e95756eee84b18e06e5f0b50342"
 dependencies = [
  "lazy-bytes-cast",
- "winapi",
+ "winapi 0.3.9",
 ]
 
 [[package]]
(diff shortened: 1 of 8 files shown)


LovecraftianHorror:
Just took a look and things are looking really _really_ great so far :tada:. This has been one of the changes I'm the most excited about, so it's great to see it getting implemented

I won't have time to dig into the code until this weekend, but just a couple of small nits / potential improvements I've noticed so far

- The reload can be a bit visually jarring. I think this is mostly due to requesting a redraw on every element pushed when we could wait until all the elements are processed before redrawing. Alternatively we could "freeze" redrawing (by setting a flag to ignore redraw events) until unfrozen
- Cache images that are fetched from urls. I would imagine local images would be more prone to tweaking when using live reloading while images from urls would be much less likely. This would help to avoid spamming a bunch of requests while quickly editing a document
- Removing the bottom of the document while looking at it can leave you scrolled past the end of the doc

All in all those points are all really minor IMO\ and I think this adds a really key feature to the viewer. Great work! :rocket: 

---

> There a few graphical glitches I'm experiencing on macOS with partial rendering

I haven't noticed any of these from the little bit I was playing with it for what it's worth

LovecraftianHorror:
I can also work on testing this on windows this weekend. I just need to setup `cargo` and keep anti-virus from silently killing things like it loves to do :upside_down_face:

trimental:
Very good points. I think I'd like the file to first load in incrementally and then in file reloads just wait until it's fully loaded. Caching url images is a very good idea, I think a hashmap to image data would suffice. Thanks for spotting the scroll_y bug, easy fix. I'll try get these done tonight.

trimental:
Alrighty that's all those points taken care of. Live reloading now is looking buttery smooth, and I've also managed to squash the macOS rendering bug.

LovecraftianHorror (approved):
I know this is still a draft PR, but everything looks good to me!

I'll probably do a refactoring pass on the interactions between the file watcher, interpreter, and event loop next next weekend

P.S. This is really really cool and still really _really_ fast. The interpreter churning through all the data of reloading `bun.md` takes ~80ms for my laptop :rocket:

PR #38, opened 2022-08-28
Custom keybindings

### LINKED ISSUE #34: Custom Keybinds
I'm planning on adding support for specifying custom keybinds this weekend. This is the general gist of how I want to expose it

```toml
# Custom keybinds for actions
# Each entry maps an action to a key combo
# Action -> predefined list of possible interactions to perform
# Key Combo -> Either a single modified key or an array of modified keys
# Modified Key -> Either an unmodified key or an obj with a key and modifiers
# Possible Modifiers: ["Alt", "Ctrl", "Os", "Shift"]
#     (Note: shift can only be used on keys that don't have a different key for
#     no-shift and shift like "Enter" or "Space")
# Possible Actions: [
#     "ToTop", "ToBottom",
#     "ScrollUp", "ScrollDown",
#     "PageUp", "PageDown",
#     "ZoomIn", "ZoomOut", "ZoomReset",
#     "Copy",
#     "Quit",
# ]
# Possible Keys: [
#     ... All the typical letters, numbers, and symbols
#     "Enter", "Space", "Tab", Backspace", "Esc", PageUp", "PageDown",
#     "PageLeft", "PageRight", "Home", "End", "Delete", "Insert", "F1"-"F12", etc.
# ]
keybinds = [
    ["ToTop", "Home"],
    ["ToBottom", "End"],
    ["Copy", { key = "c", mod = ["Ctrl"] }],
    ["ToTop", ["g", "g"]],
    ["ToBottom", "G"],
    ["Copy", "y"],
]
```

One thing that I noticed that may complicate things is that a surprising number of keys don't get picked up with a `VirtualKeyCode` on my keyboard like $%^&() etc.

Another thing is that there may be system differences here too. I noticed that @trimental uses `VirtualKeyCode::Equals`+Shift to detect a "+", but for me it registers as a `VirtualKeyCode::Plus`. It's actually impossible for me to use `VirtualKeyCode::Equals` with shift on my keyboard layout. Switching to using [the scancode](https://www.lookuptables.com/coding/keyboard-scan-codes) could fix this and the missing keys issue

LovecraftianHorror:
Resolves #34 

This implements things as listed in #34 with a couple of small tweaks due to issues in `winit`

- The amount of usable keys is slimmed down
- Shift now has to be specified regardless of the key (I originally wanted "G" to represent Shift+g, but the platform inconsistencies with how things like `+` are expressed (Shift+= vs Shift++) made me walk this back)

---

There is now a new section in the config file for keybindings. Here's the one from `inlyne.toml.sample

```toml
[keybindings]
# Base will override the defaults if set
# Default: Not set
base = [
    ["ToTop", "Home"],
    ["ToBottom", "End"],
    ["Copy", { key = "c", mod = ["Ctrl"] }],
    ["ToTop", ["g", "g"]],
    ["ToBottom", { key = "g", mod = ["Shift"] }],
    ["Copy", "y"],
]
# Extra will be applied on top of base/defaults
extra = ... # Refer to `base`
```

A decent amount of the code goes into deserializing this. I wanted to allow for the config to be simple, so you can see that you can do things like provide a single key instead of a full combo, or provide just the key if there are no modifiers

The other bulk of the changes is for handling `KeyCombos`. It's a struct that gets initialized with the keybindings. Afterwards it can consume keys and will emit if they corresponded to some action

---

Here's everything in action. The section of my config for reference

```toml
[keybindings]
extra = [
    ["Copy", "y"],
    ["ScrollUp", "k"],
    ["ScrollDown", "j"],
    ["ToTop", ["g", "g"]],
    ["ToBottom", { key = "g", mod = ["Shift"] }],
    ["Quit", "q"],
    ["Quit", [{ key = "z", mod = ["Shift"] }, { key = "z", mod = ["Shift"] }]],
    ["Quit", [{ key = "z", mod = ["Shift"] }, { key = "q", mod = ["Shift"] }]],
]
```

https://user-images.githubusercontent.com/30302768/187098807-18d806e2-fa8c-4afc-9c87-588a196a3c9a.mp4


trimental:
Looks brilliant can't wait to try this out when I review it tonight

trimental:
I tried it and I love it. The vim style shortcuts feel really good and honestly I might be tempted to implement a ghetto version of `10j, 3k..etc` in the future just to move around faster in vim style. Looking at the code its clear to see how well thought out this was, so I think this is good to merge the way it is.

With this, checklist support, customising line scrolling and live reloading, I'll make another weekly release of `v0.2`. Unless any other spectacular changes come along I think I'll change to bi-weekly releases after this one for stability. 🥳

LovecraftianHorror:
Some sort of vim-mode shouldn't be too hard to hack into the current system since it would just involve:

- Erroring on commands that start with numbers
- Special-casing if we get some numbers before a command
- Adding vim-like keybindings to the default keybindings

trimental:
Maybe you could match the number for any current bindings and if there isn't any then you store the number for the next time that you press j or k. That way you could still accept bindings starting with numbers, as I think a user would understand if they bind 5 to an action they can't expect 5j to work. About the vim-like bindings been set as default I think a lot of users would like that.

Issue #48, opened 2022-10-14
Add support for relative links


nerdachse:
Hello!

First off: Thanks for inlyne, I love it!

## Usecase

My usecase is that I have a self-contained collection of markdown files in one folder and don't want to hardcode absolute paths for easier synchronization between different clients.

## Problem

My problem is with a file like this:

```markdown
# Summary

[Introduction](README.md)

- [IT](./it/it.md)
  - [Programming Languages](./it/proglang/proglang.md)
    - [Rust](./it/proglang/rust.md)
```

When I click a link I get:
```
Error: Could not read file at "./it/proglang/proglang.md"

Caused by:
    No such file or directory (os error 2)
```

## (A hacky) Solution

```rust
if is_md {
    // Open markdown files ourselves
    let mut args = self.args.clone();
    let maybe_path = maybe_path.expect("not a path");
    // Handle relative paths and make them
    // absolute by prepending current
    // parent
    let maybe_path = if maybe_path.is_relative() {
        // Simply canonicalizing it doesn't suffice and leads to "no such file or directory"
        let current_parent = args.file_path.parent().expect("no current parent");
        let link_without_prefix: &Path = maybe_path.strip_prefix(std::path::Component::CurDir).expect("no CurDir prefix");
        let mut link = current_parent.to_path_buf();
        link.push(link_without_prefix);
        link
    } else {
        maybe_path
    };
    ...
```

PR #49, opened 2022-10-14
 #48 Add support for relative links by prefixing them with the currently

nerdachse:
opened file's parent path

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -468,7 +468,7 @@ impl Inlyne {
                                     let maybe_path = PathBuf::from_str(link).ok();
                                     let is_local_md = maybe_path.as_ref().map_or(false, |p| {
                                         p.extension().map_or(false, |ext| ext == "md")
-                                        && !p.to_str().map_or(false, |s| s.starts_with("http"))
+                                            && !p.to_str().map_or(false, |s| s.starts_with("http"))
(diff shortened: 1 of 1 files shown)


trimental:
Thanks for this @nerdachse, I'm glad you love inlyne! The only thing I would change is the http handling, at least on macOS, paths cannot be created starting with http, which is another bug that I missed. (ie opening http://markdown.md). Perhaps a nice way to handle http would be:

```rust
let is_md = maybe_path.as_ref().map_or(false, |p| {
    p.extension().map_or(false, |ext| ext == "md")
    && !p.to_str().map_or(false, |s| s.starts_with("http"))
});
```

that way its handled before the path is created. Apart from that everything looks good and runs as expected on macOS!

nerdachse:
@trimental thanks for that suggestion, it's way more elegant!
 
I incorporated it and took the liberty to rename `is_md` to `is_local_md`

Also note that your suggestion is NOT optional. With my code a http-link wouldn't open in the browser, but not at all. Now it works as expected!

trimental:
Merged. Thank you for the PR ❤️w

nerdachse:
A pleasure doing business with you, good Sir! ;)

Thanks for this awesome project!

Issue #50, opened 2022-10-14
[Suggestion] (Optionally) follow links instead of opening a new instance of inlyne


nerdachse:
Hey there!

## Motivation

Currently when you click a link to a local markdown file it spawns a new process / instance of `inlyne` with the new file.
Navigating around a hierarchy of files can create quite a few processes.

## Proposal

Make it so that clicking a local .md file replaces the currently rendered file instead.

### Optional: Middle mouse click / Shift click triggers current behavior

Maybe when clicked with the middle mouse button or with shift it opens a new instance, akin to how browsers behave (although they create a new tab instead of a new window)?

trimental:
I think this would be a great idea actually, and very simple to implement. All you have to do is check if `modifiers.shift()` is active, and if it is then do what we do normally in `is_local_md`, otherwise set `self.args.file_path` to the new path and call `event_loop_proxy.send_event(InlyneEvent::FileReload)`. 

It would make a great second PR if you feel like it 😄

nerdachse:
Oh, I do feel like it and plan working on it next week, have to hop away now!

PR #52, opened 2022-10-19
 #50 open local md files in current window when clicking - if SHIFT is

nerdachse:
held down, open them in a new window instead


trimental:
Thankfully a lot of the code needed for this was setup for the live-reloading. Looks good, I think this makes more intuitive sense then before, thanks!

PR #61, opened 2023-03-18
added page_width option to limit rendering width.

hdcommits:
Only tested it a little, seems to work.

### DIFF
--- a/inlyne.toml.sample
+++ b/inlyne.toml.sample
@@ -2,6 +2,10 @@
 # # Default: The window's scale factor (system dependent)
 # scale = 1.5
 
+# # A maximum page width can be specified
+# # Default: Unconstrained
+# page-width = 800
+
 # # The theme can be set as well
 # # Default: "Light"
 # # Possible values: ["Light", "Dark"]
--- a/inlyne.toml.sample
+++ b/inlyne.toml.sample
@@ -2,7 +2,7 @@
 # # Default: The window's scale factor (system dependent)
 # scale = 1.5
 
-# # A maximum page width can be specified
+# # A maximum page width can be specified in pixels
 # # Default: Unconstrained
 # page-width = 800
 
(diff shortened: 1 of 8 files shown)


CosmicHorrorDev:
I can give it a whirl tomorrow to make sure things seem to work fine on X11 on Linux, but at a glance the code seems good!

inlyne.toml.sample
```diff
@@ -2,6 +2,10 @@
 # # Default: The window's scale factor (system dependent)
 # scale = 1.5
 
+# # A maximum page width can be specified
```
CosmicHorrorDev, line 5:
```suggestion
# # A maximum page width can be specified in pixels
```

I think the unit is already obvious, but never hurts to clear it up

src/opts/cli.rs
```diff
@@ -62,7 +63,18 @@ pub fn command(scale_help: String, default_theme: ThemeType) -> Command {
         .value_parser(value_parser!(f32))
         .help(scale_help);
 
-    command!().arg(file_arg).arg(theme_arg).arg(scale_arg)
+    let page_width_arg = Arg::new("page_width")
+        .short('w')
+        .long("page width")
+        .number_of_values(1)
+        .value_parser(value_parser!(f32))
+        .help("Maximum width of page");
```
CosmicHorrorDev, line 71:
```suggestion
        .help("Maximum width of page in pixels");
```

Ditto

CosmicHorrorDev (commented):
Sorry for the delay

I really love these changes and everything worked how I would expect! Just a couple small nits from me

trimental:
Hey sorry it's taken week to reply. I ran the PR earlier in the week on macOS and everything looked great, no issues. Agree with the nit from Cosmic, happy to merge once that's done. Thank you for the time taken to write the PR 👍.

PR #63, opened 2023-03-24
Update actions

CosmicHorrorDev:
Each commit should be pretty self-explanatory

I went ahead and switched away from `actions-rs` actions since they've been unmaintained for a while now and have been spitting out deprecation notices due to using an old Node runtime


trimental:
Looks a lot less verbose than actions-rs. Thanks for the changes.

PR #64, opened 2023-03-29
Default to system theme

CosmicHorrorDev:
This change _should_ change the default theme to detect the system setting when possible. This is represented by a new `auto` variant

I would certainly appreciate someone else testing that this actually works. Linux's system settings are heavily non-standardized to put it nicely and the setup I use does not indicate its theme preference in a way that gets detected

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -123,6 +123,116 @@ dependencies = [
 ]
 
 [[package]]
+name = "async-broadcast"
+version = "0.5.1"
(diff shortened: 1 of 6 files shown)


CosmicHorrorDev:
Forgot to update that I used `gsettings` to set my system theme and it worked for me (although an extra data point is always nice!)

trimental:
Yep everything looks good to me on macOS. Nice improvement.

PR #65, opened 2023-03-30
Use a custom profile for building releases

CosmicHorrorDev:
This creates a custom profile intended for published releases

I toggled on fat lto because it significantly slims down the final binary for me from 18.3 MiB to 15.2 MiB both stripped (probably also speeds things up a chunk too, but it was already pretty fast anyways). I opted for a custom profile because it **significantly** slows down release builds for me (like going from 2m30s to 4m30s for a clean release build on my laptop), so this way it's easy to only use the profile for actual releases

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -46,7 +46,7 @@ jobs:
 
     - name: Build
       run: |
-        cargo build --target ${{ matrix.target }} --release --locked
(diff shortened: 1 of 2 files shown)


trimental:
Yep for CI we should just go as optimised as we can, we're not paying for CI time so might as well 😆.

PR #66, opened 2023-03-30
Add flag for generating shell completions

CosmicHorrorDev:
This adds a flag (`--gen-completions <shell>`) that generates shell completions to stdout. It also changes CI so that all completions _should_ be generated and provided up front (assuming I didn't subtly bork the release CI job)

I can confirm that it works on zsh for me

image


trimental:
Everything looks good to me. Thank you for this 👍.

PR #68, opened 2023-04-03
Don't detect system theme in tests

CosmicHorrorDev:
Followup on the whole system-theme detection for testing

This makes it so that tests always explicitly pass what theme is being used instead of having the system theme detected. Detecting the system theme in tests is obviously problematic because it leads to different behavior depending on the system being tested on

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -657,6 +657,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "ctor"
+version = "0.1.26"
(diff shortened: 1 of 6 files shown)


CosmicHorrorDev:
This should only impact test code which already passed CI, so I'm just gonna go ahead and merge

It does pull in the `pretty_assertions` dev dependency which diffs assertion output. That's a lifesaver when it comes to figuring out what actually changed when comparing near identical `Opts`

PR #69, opened 2023-04-03
Store raw image data as LZ4 compressed blobs

### LINKED ISSUE #45: Handle streaming decoding of images
Currently `inlyne` will hold the raw image data in memory. This means that physically large (as in just large width and height) images can take up _a lot_ of memory. Here is the peak memory usage when viewing `bun`'s README for reference

image

As you can see the majority of the memory usage is from holding image data in memory (77 MB when the peak is 110 MB)

I think my current preferred approach for dealing with this would be to save the remote images locally (in the user's cache dir), and perform streaming decoding to read the image on demand when it get's rendered. From my bit of testing it looks like image rendering is handled lazily, so it shouldn't involve reading the image over and over again when rendering

Other approaches could be to hold the remote image in memory and perform streaming decoding when the image gets rendered or to do streaming decoding on the image upfront and store the raw image data in some quick to decompress in-memory blob (like zstd)

The preferred approach would likely depend on how noticeable the delay is when it comes to reading off disk and decoding

---

I'm planning on handling this change (but feel free to work on it if you want)

This is the last memory-hog aspect of `inlyne`. The remaining offenders would appear to be from GPU initialization, glyph crunching, converting non RGBA8 images to RGBA8, and converting the markdown to HTML. At first glance a lot of these look like they would be out of our hands, but I'm still planning on digging in more to check anyways.

CosmicHorrorDev:
Resolves #45

(Well not quite, but it's a considerable improvement for now)

This PR changes the raw image data to be stored as LZ4 Frame Format blobs. LZ4 because it's got very fast compression/decompression speeds and the frame format because `lz4_flex`'s block format stores the output in a `Vec` allocated for the maximum possible output size which leads to a pretty large memory spike. Granted Linux (and Mac?) typically doesn't allocated zero'd pages until they're actually used for something else, but I figured the lower memory spike was worth the extra code complexity to handle streaming

## Demo

https://user-images.githubusercontent.com/30302768/229645749-dc4daaea-e092-4fba-84b2-ac4b9e84912d.mp4

### `main`

image

### This PR

image

### Analysis

Above you can see that with this PR scrolling over all of `bun`'s README is typically a steady ~20 MiB of RAM with two spikes. One towards the beginning from loading the ~55 MiB RGBA8 data before compressing it into an LZ4 blob, and a second towards the end when that blob is decompressed to get rendered

The peak memory usage is a bit lower now since the raw image data generally going to be loaded and dropped at different times. This would cause a much larger difference on a file that has more uniform sizes whereas `bun`'s README has a single outlier

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1620,6 +1620,7 @@ dependencies = [
  "image",
  "log",
  "lyon",
+ "lz4_flex",
  "memmap2",
  "notify",
  "open",
@@ -1921,6 +1922,15 @@ dependencies = [
  "float_next_after",
  "lyon_path",
  "thiserror",
+]
+
+[[package]]
+name = "lz4_flex"
+version = "0.10.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8b8c72594ac26bfd34f2d99dfced2edfaddfe8a476e3ff2ca0eb293d925c4f83"
+dependencies = [
+ "twox-hash",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,9 @@ log = "0.4.17"
 env_logger = "0.10.0"
 notify = "5.1.0"
 dark-light = "1.0.0"
+# We only decompress our own compressed data, so disable `safe-decode` and
+# `checked-decode`
+lz4_flex = { version = "0.10.0", default-features = false, features = ["frame", "safe-encode", "std"] }
 
 # Uncomment for profiling
 # [profile.release]
--- a/src/image.rs
+++ b/src/image.rs
@@ -33,6 +33,7 @@ impl ImageData {
     fn new(image: RgbaImage, scale: bool) -> Self {
         let dimensions = image.dimensions();
 
+        let start = Instant::now();
         let mut frame_info = FrameInfo::new();
         // This seems to speed up decompressing considerably
         frame_info.block_size = BlockSize::Max256KB;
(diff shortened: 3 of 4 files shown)


trimental:
Hello once again sorry for delays in checking this out. Brilliant work, everything looks smooth on my computer and the compression causes great memory savings 🥳. Just two little nitpicks.

First I think it would make more sense to only log the time of decompression, and to also print the time & data sizes when we compress. Compression takes longer so it's the more important metric to log.

Secondly I think we should preallocate the vector the same way that the lz4_flex crate does it. They use a private module function :(. But it's such a simple function we can just inlyne it to our code. Might save a few vector reallocations on large files.

` let mut lz4_enc = FrameEncoder::with_frame_info(frame_info, Vec::with_capacity(16 + 4 + (image.len() as f64 * 1.1) as usize));`

CosmicHorrorDev:
> First I think it would make more sense to only log the time of decompression, and to also print the time & data sizes when we compress. Compression takes longer so it's the more important metric to log.

That sounds good!

> Secondly I think we should preallocate the vector the same way that the lz4_flex crate does it.

That's the same technique that the lz4 block format uses (aka `lz4_flex::block`). I opted for streaming and eating a few reallocations since the compressed sizes are so small that it won't really end up resizing too many times anyways (it uses an internal buffer of 256 KiB with the current frame info settings)

We could allocate everything up front and then shrink to fit the compressed size, but that would generally end up allocating like 50x the amount of memory we need in the end. I chatted with the crate author and they said that allocating everything up front was only intended for small sizes (like sub 8 MiB), and to use the frame format when dealing with anything larger

---

I'll work on tweaking the logs later today. Thanks for the review as always :)

PR #71, opened 2023-04-06
Make everything async

trimental:
Use Tokio to make everything async instead of using threads. This mainly impacts image loading which used to use thread spacing before. I feel as though this is a bit cleaner at the cost of feeling a bit 'magic-like'. I'm interested to hear your thoughts about this @CosmicHorrorDev. 

I haven't done any benchmarking against the old implementation yet. This should help with streaming images directly into a compressed format.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -199,6 +199,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "async-once-cell"
+version = "0.4.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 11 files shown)


CosmicHorrorDev:
I'll have to take a closer look this weekend, but I'm interested in switching things over to async since it seems to be a better fit for UIs in general. Exciting :smile:

CosmicHorrorDev:
Since this has some overlap with #69 I'm okay with handling getting this merged first and then I can rebase that branch on top btw

trimental:
I actually did this branch over #69 initially and just seperated them. I'll merge yours tonight and then rebase this since I already have the code for it.

PR #75, opened 2023-04-14
Use srgb colors only when on a surface texture format that uses srgb

trimental:
Currently we've been using srgb everywhere which leads to some funny color differences between platforms and potentially devices. This pr only converts themes to srgb if using a srgb surface format. For now I've only put the two most common srgb formats. If you run into any device that uses anything else please let me know.


CosmicHorrorDev:
Looks good from a brief skim. I can take a closer look tomorrow!

CosmicHorrorDev (approved):
Everything looks fine to me and the code looks much cleaner now :+1:

PR #76, opened 2023-04-18
Add a `deny.toml` file for `cargo-deny`

CosmicHorrorDev:
This just adds portions for dealing with the licenses of all of our dependencies. `cargo-deny` also supports a lot more functionality on dealing with security advisories or detecting duplicate dependencies that could be useful as well, but wasn't the focus of this PR

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -1,0 +1,26 @@
+[licenses]
+allow = [
+    "MIT",
+
+    "Apache-2.0",
+    "BSD-2-Clause",
+    "BSD-3-Clause",
+    "ISC",
+    "MPL-2.0",
+    "Zlib",
+]
+
+exceptions = [
+    # We will no longer depend on this dependency after `copypasta` updates
+    # `clipboard-win`. Keeping it separate as a reminder to drop eventually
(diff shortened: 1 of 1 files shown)


trimental:
Hmm I'm a bit hesitant about this. I'm a very copy left person. So copy left that I don't even really believe in licenses. If you can physically see and view a piece of code, bar actually breaking into someones machine to steal it, I think you should be able to use it. 

Same way that mathematicians that publish their formulas cannot copyright them. I only include the MIT license to satisfy people that do care about licenses. Like you said in the comment, so many licenses are being broken every day with no outcome its comic.

But on a less philosophical note maybe this would be a little useful but I can't really justify using a cargo plugin for something that could just be a PR to remove a line from Cargo.toml. Thanks for this idea though, happy to discuss further :)

CosmicHorrorDev:
I'm fine either way. To me licenses represent how downstream authors want their code to be handled, so it seems respectful at the very least to try and follow that

If we don't merge this then I'll probably just include it locally to keep an eye on things from time to time, so it doesn't really matter to me if it's actually committed or not

CosmicHorrorDev:
Not a big deal here, so I'm gonna go ahead and close this

PR #77, opened 2023-04-23
Go back to using threads to spawn images, removing async

trimental:
Switch back to threads due to performance challenges with async. Hopefully this should still perform better then before due to queuing up reposition events when image load in, as well as improving removing the logic to manually add a callback to the image loading thread.


README.md
```diff
@@ -1,7 +1,6 @@
 <h1 align="center">Inlyne - a GPU powered, browserless, markdown + html viewer</h1>
 
 <p align="center">
-<a href="https://crates.io/crates/inlyne" target="_blank"><img height=25 src="https://img.shields.io/crates/v/inlyne.svg" /></a>
```
CosmicHorrorDev, line 4: Was this supposed to be removed?

README.md
```diff
@@ -45,7 +44,6 @@ the document where you left off. It's designed to work seamlessly and allow you
 | Favourite band    | Nickleback         |
 
 #### Sizable images
-<img src="https://i.redd.it/hwurhp7crzf81.png" width = "300" align="center"></img>
```
CosmicHorrorDev, line 48: Ditto with above

CosmicHorrorDev (commented):
Everything looks great to me! Just a couple of small comments :+1:

README.md
trimental: Yes certainly not suppose to be removed oops. Was just testing if the presence of images induced any blocking. I'll fix this up thanks.

PR #85, opened 2023-05-07
Temporarily skip generating completions for aarch64-apple-darwin release

CosmicHorrorDev:
For some reason this is failing in CI with a bad cpu type

In the future we generate these static assets before-hand or have them generated from a build script instead of generating them at run-time in the release pipeline

Would be good to repush the release tag after this is merged

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -67,11 +67,14 @@ jobs:
           inlyne_bin="$staging/inlyne"
         fi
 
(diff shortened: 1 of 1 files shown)


trimental:
Yep looks good. We can sort this out in the future. I'll repush the release. Thank you.

trimental:
Alright fingers crossed, lets see how this goes this time.

Issue #84, opened 2023-05-07
Allow for flag to overwrite config location


coastalwhite:
It would be very useful to have a `--config` flag to select a custom configuration file.

CosmicHorrorDev:
I would be okay with this (along with something like `--no-config` to easily skip using a config file)

Just to get perspective for workflows, I'm curious if you have some specific motivating use-case for this

coastalwhite:
This is very useful to alias `inlyne` to `inlyne --config <config.toml>` in case you want to use a separate environment for a while. This would also be useful, for example, if you wanted to embed this program into another program.

trimental:
@CosmicHorrorDev do you mean --no-config to skip using the config from the default location? Otherwise I'm guessing you just wouldn't put a --config flag.

CosmicHorrorDev:
`--no-config` to skip reading config files altogether (e.g. how `ripgrep` uses it). I've seen some other names for it like `--clean` for vim

It's useful for making sure you get a clean environment for things like bug reports

trimental:
Might be useful for testing

CosmicHorrorDev:
@coastalwhite feel free to work on this if you want to btw. I've got other stuff I need to get through today, so I won't be able to start on it till sometime next week

PR #91, opened 2023-05-08
Some clean ups, fixes, config flag

coastalwhite:
This PR cleans up some aspects of the configuration management. It also implements #84. This also fixes the long form `page width` flag, which was previously broken.

### DIFF
--- a/src/keybindings/defaults.rs
+++ b/src/keybindings/defaults.rs
@@ -4,98 +4,100 @@ use winit::event::{ModifiersState, VirtualKeyCode as VirtKey};
 
 const IS_MACOS: bool = cfg!(target_os = "macos");
 
-pub fn defaults() -> Keybindings {
-    let ctrl_or_command = if IS_MACOS {
-        ModifiersState::LOGO
-    } else {
-        ModifiersState::CTRL
-    };
+impl Default for Keybindings {
+    fn default() -> Self {
+        let ctrl_or_command = if IS_MACOS {
+            ModifiersState::LOGO
+        } else {
+            ModifiersState::CTRL
+        };
 
-    vec![
-        // Copy: Ctrl+C / Command+C
-        (
-            Action::Copy,
-            KeyCombo(vec![ModifiedKey(Key::from(VirtKey::C), ctrl_or_command)]),
-        ),
-        // Zoom in: Ctrl++ / Command++
-        (
-            Action::ZoomIn,
-            KeyCombo(vec![ModifiedKey(
-                Key::from(VirtKey::Equals),
-                ctrl_or_command | ModifiersState::SHIFT,
-            )]),
-        ),
-        (
-            Action::ZoomIn,
-            KeyCombo(vec![ModifiedKey(
-                Key::from(VirtKey::Plus),
-                ctrl_or_command | ModifiersState::SHIFT,
-            )]),
-        ),
-        // Zoom out: Ctrl+- / Command+-
-        (
-            Action::ZoomOut,
-            KeyCombo(vec![ModifiedKey(
-                Key::from(VirtKey::Minus),
-                ctrl_or_command,
-            )]),
-        ),
-        // Zoom reset: Ctrl+= / Command+=
-        (
-            Action::ZoomReset,
-            KeyCombo(vec![ModifiedKey(
-                Key::from(VirtKey::Equals),
-                ctrl_or_command,
-            )]),
-        ),
-        // Scroll up: Up-arrow
-        (Action::ScrollUp, KeyCombo::from(VirtKey::Up)),
-        // Scroll down: Down-arrow
-        (Action::ScrollDown, KeyCombo::from(VirtKey::Down)),
-        // Go to top of doc: Home
-        (Action::ToTop, KeyCombo::from(VirtKey::Home)),
-        // Go to bottom of doc: End
-        (Action::ToBottom, KeyCombo::from(VirtKey::End)),
-        // Quit: Esc
-        (Action::Quit, KeyCombo::from(VirtKey::Escape)),
-        // vim-like bindings
-        // Copy: y
-        (Action::Copy, KeyCombo::from(VirtKey::Y)),
-        // Scroll up: k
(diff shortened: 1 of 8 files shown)


trimental:
This changes look good to me. I'll let @CosmicHorrorDev give the final review since this mainly touches upon the code he has written.

CosmicHorrorDev:
I'll give it a look after I'm off work!

src/keybindings/defaults.rs
```diff
@@ -4,98 +4,100 @@ use winit::event::{ModifiersState, VirtualKeyCode as VirtKey};
...
-    } else {
-        ModifiersState::CTRL
-    };
+impl Default for Keybindings {
+    fn default() -> Self {
+        let ctrl_or_command = if IS_MACOS {
+            ModifiersState::LOGO
+        } else {
+            ModifiersState::CTRL
+        };
```
CosmicHorrorDev, line 13: I think I'd prefer something like `pub fn defaults() -> Vec<(Action, KeyCombo)>` in this file still and then

```rust
impl Default for Keybindings {
    fn default() -> Self {
        Self(defaults::defaults())
    }
}
```

in mod.rs just to try and keep impls contained in the same file

I definitely like making `Keybindings` into a proper type though!

src/keybindings/mod.rs
```diff
@@ -4,21 +4,42 @@ mod serialization;
...
+pub struct Keybindings(Vec<(Action, KeyCombo)>);
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
 pub enum Key {
     Resolved(VirtualKeyCode),
     ScanCode(ScanCode),
 }
 
+impl Keybindings {
+    #[inline]
```
CosmicHorrorDev, line 22:
```suggestion
```

Nit: Slapping `#[inline]` on functions without justifying through  something like checking the asm or benchmarks is generally a code smell. Chances are LLVM will already inline this and even if it doesn't this function isn't called nearly enough to be perf sensitive

src/opts/config.rs
```diff
@@ -83,15 +90,25 @@ pub struct Config {
 }
 
 impl Config {
-    pub fn load() -> anyhow::Result<Self> {
-        let config_dir = dirs::config_dir().context("Failed detecting config dir")?;
+    pub fn load_from_file(path: &Path) -> anyhow::Result<Self> {
+        let config_content = read_to_string(path).context(format!(
+            "Failed to read configuration file at '{path}'",
+            path = path.display()
+        ))?;
```
CosmicHorrorDev, line 97:
```suggestion
        let config_content = read_to_string(path).context(format!(
            "Failed to read configuration file at '{}'",
            path.display(),
        ))?;
```

I think it's clearer to just use the typical interpolation style here

src/opts/mod.rs
```diff
@@ -76,63 +73,76 @@ impl Opts {
...
-            lines_to_scroll: config_lines_to_scroll,
-            light_theme: config_light_theme,
-            dark_theme: config_dark_theme,
-            font_options: config_font_options,
-            keybindings:
-                config::KeybindingsSection {
-                    base: keybindings_base,
-                    extra: keybindings_extra,
-                },
-        } = config;
```
CosmicHorrorDev, line 100: I prefer destructuring all the values here as a way of future proofing that any additions to `Config` will be a hard compiler error here. Sure it's more code, but it also helps guide the programmer on what needs to be changed

It really should be done for `args` too. Not sure why I didn't

src/opts/cli.rs
```diff
@@ -43,32 +47,36 @@ pub fn command(scale_help: String, default_theme: Option<ThemeType>) -> Command
...
 
-    let page_width_arg = Arg::new("page_width")
+    let config_arg = Arg::new("config")
+        .short('c')
+        .long("config")
+        .number_of_values(1)
+        .value_parser(value_parser!(PathBuf))
+        .help("Configuration file to use");
+
+    let page_width_arg = Arg::new("page width")
```
CosmicHorrorDev, line 77:
```suggestion
    let page_width_arg = Arg::new("page_width")
```

The arg reads a little awkward in the help text with the space in the center (`-w, --page-width <page width>`)

src/opts/cli.rs
```diff
@@ -78,60 +86,38 @@ pub fn command(scale_help: String, default_theme: Option<ThemeType>) -> Command
...
             Self::print_completions(shell, &mut c);
             std::process::exit(0);
         }
 
         let file_path = matches.get_one("file").cloned().expect("required");
         let theme = matches.get_one("theme").cloned();
         let scale = matches.get_one("scale").cloned();
-        let page_width = matches.get_one("page_width").cloned();
+        let config = matches.get_one("config").cloned();
+        let page_width = matches.get_one("page width").cloned();
```
CosmicHorrorDev, line 114:
```suggestion
        let page_width = matches.get_one("page_width").cloned();
```

Other part of the last comment

CosmicHorrorDev (changes requested):
Just a few small nits, but otherwise looks great :+1:

This looks like three changes in a trenchcoat, so I'll address each in turn

---

**Fixing the broken page width flag**

Yikes, good find! I'll add a regression test to make sure that it actually gets tested for parsing in a follow-up PR

---

**Various cleanups**

I really appreciate all the changes. Everything seemed great to me!

---

**Adding a `--config` flag**

I see that this dropped modifying default values for `clap`'s help text with values from the config. Technically we _could_ still manage it by trying to parse args -> reading the config file -> actually parse the args, but a) I don't think that's worth the effort and b) I don't even think the original implementation before this PR was worth the effort. It gets a goodbye and good riddance from me :grin:

---

Overall I love the changes! Thanks for the PR

coastalwhite:
I will have a look at this tonight

CosmicHorrorDev:
I'll take a look when I'm on my lunch break in about half an hour

CosmicHorrorDev (approved):
Thanks for all the changes, everything looks great!

PR #98, opened 2023-05-09
Fix --page-width flag

figsoda:


CosmicHorrorDev (approved):
This also gets fixed as a part of #91, but I'm happy to merge this PR on its own since #91 has a lot more going on. Thanks!

PR #99, opened 2023-05-09
Release v0.3.1

CosmicHorrorDev:
This is a boring release to just get the `--page-width` fix out

This commit is tagged with v0.3.1 which I can push after this is merged

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "inlyne"
-version = "0.3.0"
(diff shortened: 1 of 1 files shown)


trimental (approved):

CosmicHorrorDev:
@trimental I see now that the release pipeline doesn't publish to crates. Looks like you'll have to publish the current `main` (alternatively you can add me as a crate owner and I can publish after I get off work. My lunch break is over)

PR #100, opened 2023-05-09
Update dependencies

CosmicHorrorDev:
Just a `cargo update`. Should have done this with the initial release. The lock file being outdated caused a build failure since it was building with the `--locked` flag

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "ab_glyph"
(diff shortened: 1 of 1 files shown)


trimental:
This won't let me release v3.0.1 because of https://github.com/CosmicHorrorDev/inlyne/blob/0a1193f2ca862db7c52ffc58ae63f1269d600063/Cargo.lock#L1676 which should be v3.0.1

PR #101, opened 2023-05-09
Bump inlyne version in the lockfile

CosmicHorrorDev:
See https://github.com/trimental/inlyne/pull/100#issuecomment-1540818432


trimental:
I've just run cargo update again since it pulled in one more update as well. I've also sent you an invite for crate ownership for the future 👍

PR #113, opened 2023-05-14
Generate shell completions through a `cargo-xtask` task

CosmicHorrorDev:
This changes generating completions from a runtime flag to being generated ahead of time using a [`cargo-xtask`](https://github.com/matklad/cargo-xtask) task

This resolves the following workaround in CI

```bash
# Skip `aarch64-apple-darwin` for now since it's failing in CI
```

along with avoiding pulling in `clap_complete` as an `inlyne` dependency

Generating shell completions is done by running

```bash
$ cargo xtask gen
```

and sets up the framework for other similar tasks (such as generating a man page)

The major downside is that this does make generating the content a manual process, but we can just follow `rust-analyzer`'s technique and handle releases through an `xtask` too. This would make it easy to either check or generate static content before publishing the release

_(Note: since this removes a command line flag the next release will have to be a semver breaking version bump)_


trimental:
Finally got the chance to look through this. I'm no expert when it comes to things like cargo xtask but it seems everything works correctly for me. Thanks for writing all this. +337 lines is no small undertaking.

trimental (approved):

PR #114, opened 2023-05-14
Rename sample TOML file

CosmicHorrorDev:
Renames `inlyne.toml.sample` -> `inlyne.sample.toml` and removes one layer of comments

The main motivation here is to ensure that syntax highlighting and editor support works well since it is now a true TOML file, and acts as a proper sample config instead of a large block on comments. I figure this is more of what people would expect for a sample config file.

We could also setup a default config file that's meant to be a jumping off point for users. Something like an `inlyne.defaults.toml`. It could also make sense to provide only an `inlyne.defaults.toml` file instead


trimental:
Hmm yeah a inlyne.defaults.toml makes sense to me. I feel as though that could probably just replace the sample.toml. But for now that is a good fix for a hindsight.

PR #117, opened 2023-05-15
Support parsing shortcodes

CosmicHorrorDev:
This adds support for parsing emoji shortcodes like `:tada:`

Should probably wait until after #105 is merged, so that rendering emojis is in better shape


CosmicHorrorDev:
Tested with changes from current `main` and this is working as expected :+1:

PR #120, opened 2023-05-20
Fork glyphon and cosmic-text

trimental:
Two major PRs we're waiting on are https://github.com/grovesNL/glyphon/pull/34 and https://github.com/pop-os/cosmic-text/pull/132. That implement text texture atlas growth and monospace fallback respectively.

Looking at the activity of both crates it might be a while until these can make it into the codebase. Will probably be a good idea to switch back once these two PR's are merged.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -675,8 +675,7 @@ dependencies = [
 [[package]]
 name = "cosmic-text"
 version = "0.8.0"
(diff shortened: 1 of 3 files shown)


CosmicHorrorDev (approved):
Seems to fix both of the issues for me :+1:

I've subscribed to both PRs, so we'll see whenever they get feedback

PR #121, opened 2023-05-20
Default light theme to modified `InspiredGitHub`

CosmicHorrorDev:
Looks like we both ended up making pretty similar changes :laughing: 

This switches the default theme to `InspiredGitHub`, adds support for `font-style` since the code highlighting also makes use of that, and overrides the background color to match the background color used by GitHub

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -546,7 +546,7 @@ dependencies = [
  "heck",
  "proc-macro2",
  "quote",
- "syn 2.0.15",
+ "syn 2.0.16",
 ]
 
 [[package]]
--- a/src/color.rs
+++ b/src/color.rs
@@ -57,7 +57,7 @@ pub const LIGHT_DEFAULT: Theme = Theme {
     link_color: 0x5466FF,
     select_color: 0xCDE8F0,
     checkbox_color: 0x96ECAE,
-    code_highlighter: SyntaxTheme::Base16OceanLight,
+    code_highlighter: SyntaxTheme::InspiredGithub,
 };
 
 #[derive(Deserialize, Clone, Copy, Debug, PartialEq, Eq)]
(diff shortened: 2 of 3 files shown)


CosmicHorrorDev:
Oh I also noticed that italic doesn't seem to be working after the glyphon switch. Even trying something simple like

```markdown
_Italic_
```

Renders with normal text for me

trimental:
Hmm I don't know about changing the default 

> Oh I also noticed that italic doesn't seem to be working after the glyphon switch. Even trying something simple like
> 
> ```gfm
> _Italic_
> ```
> 
> Renders with normal text for me

Oops going to push a fix for this.

CosmicHorrorDev:
I'm fine with reverting that part of the change. There's no one code highlighter that I'm 100% happy with unfortunately

trimental:
Alright italics fixed e47443e02b65330a25f9dc33309e520fdefcf7c3. I like the change to github inspired but I'm unsure about whether we need to change the background. Why did you remove the change to InspiredGithub?

CosmicHorrorDev:
> I like the change to github inspired but I'm unsure about whether we need to change the background

I personally think the background should be switched otherwise we end up with the same background color for both code blocks and the regular background by default. The value I picked is what github uses for it's light theme which looks much better IMO

image

> Why did you remove the change to InspiredGithub?

I must have misunderstood what you were saying with "Hmm I don't know about changing the default" since I thought you were talking about changing the default theme

CosmicHorrorDev:
Alright, reverted back to changing the default syntax theme to InspiredGitHub

I would prefer sticking with the change to code blocks background color. I don't know of a markdown viewer that doesn't distinguish code blocks with a different background or some outline at the very least

trimental:
Yeah I see your point, maybe we should try and get that change upstreamed to syntec if we can. For now the hacks ok, I just hope no one else sets that css property to 0xFFFFFF in html for something else and is surprised but I guess thats a low chance.

trimental (approved):

trimental:
> I must have misunderstood what you were saying with "Hmm I don't know about changing the default" since I thought you were talking about changing the default theme

Lol meant to type 'default background'

CosmicHorrorDev:
I'll look into fiddling with the syntect plugin for comrak to see if we can start getting more control over the code highlighting

PR #122, opened 2023-05-20
Make correct location of `inlyne.toml` file clearer

nicoburns:
I had to look up the `dirs` documentation to work out where the config directory was on macOS, and I had to look through the inlyne source code to determine that the file was nested under a directory called `inlyne`.


CosmicHorrorDev (approved):
Definitely worth pointing out :+1:

I'll merge once CI finishes

PR #123, opened 2023-05-21
Properly modify InspiredGitHub's background color

CosmicHorrorDev:
This properly handles modifying Inspired GitHub's background color without the hack in the interpreter

The themes in a `ThemeSet` are just stored in `BTreeMap`, so we could always choose to load in our own default `inlyne-light` and `inlyne-dark` themes if desired

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1652,6 +1652,7 @@ dependencies = [
  "resvg",
  "serde",
  "serde_yaml",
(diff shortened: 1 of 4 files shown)


trimental:
Ooo very clean fix thanks for looking into that. Much happier with this.

trimental (approved):

PR #124, opened 2023-05-21
`inlyne.sample.toml` -> `inlyne.default.toml`

CosmicHorrorDev:
This switches the provided `inlyne.sample.toml` file to the `inlyne.default.toml` file filled with default values*

The two main benefits here are

1. We now have a test that ensures the default values stay in sync instead of relying on people to update the comments
2. Users can copy the whole `inlyne.default.toml` file and start tweaking values in it instead of having to start from fresh with the `inlyne.sample.toml` file as a reference


nicoburns:
It would be cool if:

- The file was automatically created (no copying neccessary)
- And ideally if there was a keyboard shortcut or some other way to open to file (or at least view it's path)

CosmicHorrorDev:
Maybe something like an `--edit-config` flag that creates the file if it doesn't exist and opens it with the user's `$EDITOR` (well for Mac/Linux at least. Windows probably has some other mechanism for opening a file with the default app)

trimental (approved):

trimental:
Looks good. Fits with how alacritty handles things as well.

PR #126, opened 2023-05-21
Anchorize footnotes

### LINKED ISSUE #90: Extended Syntax - Footnotes would be very nice.
Would be excellent to support Extended Syntax, expecially [Footnotes](https://www.markdownguide.org/extended-syntax/#footnotes).

Like this[^1].

```markdown
Like this[^1].

[^1]: A [footnote](https://www.markdownguide.org/extended-syntax/#footnotes)
```

[^1]: A [footnote](https://www.markdownguide.org/extended-syntax/#footnotes)

CosmicHorrorDev:
Resolves #90

A footnote will will generate an `"id"` on both the hyperlink on the main text and the list item on the footnote that both serve as anchors. The generated html is as follows

```html
<p>Like this<sup class="footnote-ref"><a href="#fn-1" id="fnref-1" data-footnote-ref>1</a></sup>.</p>
<section class="footnotes" data-footnotes>
    <ol>
        <li id="fn-1">
            <p>A <a href="https://www.markdownguide.org/extended-syntax/#footnotes">footnote</a> <a href="#fnref-1" class="footnote-backref" data-footnote-backref aria-label="Back to content">↩</a></p>
        </li>
    </ol>
</section>
```

Where in the above `#fn-1` is the anchor for the main text and `#fnref-1` is the anchor for the footnote

The footnote on the list item is a little awkward since the anchor is positioned in `Positioner::position()` which doesn't get sent list items. Instead a pending anchor is pushed that gets attached to the inner paragraph element

_Aside: The footnote doesn't get a numbered list even though it should. I'm assuming some of the more complicated elements in the footnote are causing the interpreter state to get messed up_

_Other aside: The footnote number isn't a superscript, but I think you mentioned that was one of the side effects of the glyphon switch_

## Demo

```md
# Header

Like this[^1].

1. filler
1. filler
1. filler
1. filler
1. filler
1. filler
1. filler
1. filler
1. filler
1. filler
1. filler
1. filler
1. filler
1. filler
1. filler

Another footnote[^2].

1. filler
1. filler
1. filler
1. filler
1. filler
1. filler
1. filler
1. filler
1. filler
1. filler
1. filler
1. filler
1. filler
1. filler
1. filler

- [To Header](#header)

[^1]: A [footnote](https://www.markdownguide.org/extended-syntax/#footnotes)
[^2]: 2nd [footnote](https://www.markdownguide.org/extended-syntax/#footnotes)
```

https://github.com/trimental/inlyne/assets/30302768/704eb582-f882-4bba-a8d2-9cff80759c1f

### DIFF
--- a/src/interpreter.rs
+++ b/src/interpreter.rs
@@ -147,6 +147,7 @@ struct State {
     span_style: FontStyle,
     // Stores the row and a counter of newlines after each image
     inline_images: Option<(Row, usize)>,
+    pending_anchor: Option<String>,
 }
 
 pub struct HtmlInterpreter {
@@ -332,11 +333,13 @@ impl TokenSink for HtmlInterpreter {
                                 .push(html::Element::Table(Table::new()));
                         }
                         "a" => {
-                            let attrs = tag.attrs;
-                            for attr in attrs {
-                                if attr.name.local == local_name!("href") {
-                                    self.state.text_options.link.push(attr.value.to_string());
-                                    break;
+                            for Attribute { name, value } in tag.attrs {
+                                if name.local == local_name!("href") {
+                                    self.state.text_options.link.push(value.to_string());
+                                }
+                                if name.local == local_name!("id") {
+                                    let anchor_name = format!("#{}", value);
+                                    self.current_textbox.set_anchor(Some(anchor_name));
                                 }
                             }
                         }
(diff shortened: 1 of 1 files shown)


trimental:
Looks like a good change. I want to focus on multisized font lines in glyphon for things like `<sup>` and `<small>` soon but I think I'll do that once I've got taffy up and running as that may be hard enough to do without adding weird positioning hacks to the text rendering code.

trimental (approved):

CosmicHorrorDev:
That'd be great! (but also no rush at all of course)

PR #127, opened 2023-05-23
Add support for custom syntax themes

CosmicHorrorDev:
The recent syntect plugin work had me thinking about how much work it would be to allow for custom themes outside of the ones bundled with syntect. Turns out the answer is not _too_ much, but there's definitely some changes I want to submit upstream for `comrak` and `syntect`

This allows for specifying custom themes by passing a `.tmTheme` file like so

```toml
[dark-theme]
code-highlighter.path = "/path/to/theme.tmTheme"
```

while still supporting the old method for specifying any of the bundled themes. There was a lot of reworking to pass around the actual syntect `Theme` instead of the enum specifying which bundled theme to use, but other than that the changes were pretty simple

## Demo

When using a slightly modified (this theme could use a lot more work though):

https://github.com/rose-pine/sublime-text/blob/main/rose-pine.tmTheme

image

### DIFF
--- a/inlyne.default.toml
+++ b/inlyne.default.toml
@@ -45,6 +45,9 @@ checkbox-color = 0x0a5301
 #     "base16-ocean-light", "inspired-github",      "solarized-dark",
 #     "solarized-light"
 # ]
+# You can also pass a path to a `.tmTheme` file for a custom theme instead
+# Example:
+# code-highlighter.path = "/path/to/custom.tmTheme"
 code-highlighter = "base16-ocean-dark"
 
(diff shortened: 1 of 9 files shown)


trimental:
Wow great code as always and a great feature 🥳. 

-------
Sorry it took me 2 days to comment on a PR this exciting but I've been trying *partially unsuccessfully* to get taffy running with my zero knowledge of css grids.

trimental (approved):

CosmicHorrorDev:
No worries! I've just been building my own binaries for local use since we probably won't have a release for a while

PR #128, opened 2023-05-26
Fix footnote numbered list

CosmicHorrorDev:
_Probably best to review the two commits separately. This probably should have been two PRs, but here we are :sweat_smile:_

The issue was that the code for pushing the list prefix would only check the last element pushed in the `element_stack` when processing `CharacterTokens(_)`, but the footnote reference generates the following HTML

```html
<section class="footnotes" data-footnotes>
    <ol>
        <li id="fn-1">
            <p>A <a href="https://www.markdownguide.org/extended-syntax/#footnotes">footnote</a> <a href="#fnref-1" class="footnote-backref" data-footnote-backref aria-label="Back to content">↩</a></p>
        </li>
    </ol>
</section>
```

which means that a `ListItem` is pushed on followed by a `Paragraph(_)` before hitting any character tokens

This PR changes things to set a `pending_list_prefix` when processing an opening `"li"` element and then checking if `pending_list_prefix` is set in the appropriate place when processing `CharacterTokens(_)` which should handle nesting better. The first commit is just some twiddling to make `HtmlInterpreter` testable since it's a big bundle of mutable state, and I don't want to accidentally regress something. A fun part of this is that the second commit shows how the list prefix is now correctly getting pushed onto the footnote reference

```diff
diff --git a/src/snapshots/inlyne__interpreter__tests__lists_get_prefixed.snap b/src/snapshots/inlyne__interpreter__tests__lists_get_prefixed.snap
index d4e5391..f27b437 100644
--- a/src/snapshots/inlyne__interpreter__tests__lists_get_prefixed.snap
+++ b/src/snapshots/inlyne__interpreter__tests__lists_get_prefixed.snap
@@ -75,6 +75,23 @@ expression: finished_queue
             indent: 50.0,
             font_size: 16.0,
             texts: [
+                Text {
+                    text: "1. ",
+                    color: None,
+                    link: None,
+                    is_bold: true,
+                    is_italic: false,
+                    is_underlined: false,
+                    is_striked: false,
+                    font_family: SansSerif,
+                    hidpi_scale: 1.0,
+                    default_color: [
+                        0.0,
+                        0.0,
+                        0.0,
+                        1.0,
+                    ],
+                },
                 Text {
                     text: "First footnote ",
                     color: None,
```

_Side note: The single test added with this PR bumps the line test coverage from 25% to 33% :tada:_

## Demo

```markdown
A footnote[^1]

- Unordered list
- Unordered list

Another footnote[^2]

1. Ordered list
1. Ordered list

[^1]: 1st footnote
[^2]: 2nd footnote
```

image

It looks like a spacer gets pushed between the footnotes, but that also happens on `main`, so it's not a regression from this PR at least

### DIFF
--- a/src/interpreter.rs
+++ b/src/interpreter.rs
@@ -93,7 +93,6 @@ mod html {
 
     pub enum Element {
         List(List),
-        ListItem,
         Input,
         Table(Table),
         TableRow(Vec<TextBox>),
@@ -148,6 +147,7 @@ struct State {
     // Stores the row and a counter of newlines after each image
     inline_images: Option<(Row, usize)>,
     pending_anchor: Option<String>,
+    pending_list_prefix: Option<String>,
 }
 
 pub struct HtmlInterpreter {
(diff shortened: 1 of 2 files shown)


CosmicHorrorDev:
On second thought I ripped out the work to make the interpreter testable. There's still more that I want to do, and I don't want to blow up the size of this PR by pulling all of that in. Instead it will be coming soon (aka this weekend)

trimental:
> since it's a big bundle of mutable state

I find this very offensive. Interpreter.rs is 900 lines of finite state machine goodness that blesses those who gaze upon it. In all seriousness though sometimes I'm afraid to touch it and break something. If anything needs testing its that.

On windows at least it seems like anchors aren't being called because `open::that` will actually accept an anchor like `#ref1`. Easy fix though, we just need to switch positions between the if and else statements for `open::that`.

That spacing is actually true to the html. If you print the htmlified output and put it in firefox, it gives:

image

CosmicHorrorDev:
Good catch on window's! I should probably setup a test environment there

Feel free to push a fix if you want

trimental:
@CosmicHorrorDev yep I will tomorrow, got a few other bug fixes I need to do as well. Seems like single glyph links broke with the glyphon PR but I know why which is good.

CosmicHorrorDev:
Awesome! I really appreciate the review

CosmicHorrorDev:
The fix looks good to me :+1:

PR #129, opened 2023-05-26
Use taffy for tables

trimental:
Really excited for this PR. Should make future additions to table layout logic a lot simpler to deal with (for images and nesting etc). 

Right now this doesn't cache layouts and forces `text_cache` and `font_system` to become an `Arc<Mutex<>>` but why fight the borrow checker when you can let it have its way with you. Plus *premature optimization is the root of all evil* so let's just enjoy properly sized tables.

Big thanks to @nicoburns for his help with this.


CosmicHorrorDev:
I'll be free to give this a look tomorrow if ya want 👍

CosmicHorrorDev (approved):
Only a couple tiny nits that can just have the suggestions approved to make things easy. Other than that everything looks good :)

Having to wrap the text cache and font system in a `Mutex` is a bit unfortunate since we don't ever try to access either simultaneously, but it looks like `Measurable` requires a `Sync` bound, so we can't use a `RefCell`. At the very least locking doesn't even show up when I profile things (probably because there's never any contention)

We could switch to `parking_lot::Mutex` just to avoid having to `.unwrap()` every time we acquire the lock. It's already in our dep tree in several places, so it doesn't actually add a dependency

nicoburns (approved):
This looks good to me. There are some refactors that could be done, but if Taffy's scope within Inlyne is going to be expanded in future (e.g. to do the main vertical layout of the page) then it probably makes more sense to make those changes then.

nicoburns:
> Right now this doesn't cache layouts

This should be quite easy to do. Just keep the `Taffy` instance around and reuse it (this would work particularly well for things like resizing where you wouldn't need to update any of the styles and could just pass in different top-level bounds). I imagine a typical markdown file is small enough that it'll be fast without any caching though :)

> Having to wrap the text cache and font system in a Mutex is a bit unfortunate since we don't ever try to access either simultaneously, but it looks like Measurable requires a Sync bound, so we can't use a RefCell. At the very least locking doesn't even show up when I profile things (probably because there's never any contention)

This is definitely fixable in Taffy. It just requires us to be generic over the `Measurable` type and some juggling of type bounds (because some people need the Taffy struct to be Send+Sync). I've put up a PR in that direction here https://github.com/DioxusLabs/taffy/pull/490

Issue #55, opened 2022-11-27
Add Wayland support


CRAG666:
His project looks great, but I can't use it on Wayland

trimental:
Wayland support should work automatically. I'm afraid I'll need more details before I can fix your specific issue.

CRAG666:
@trimental 
![imagen](https://user-images.githubusercontent.com/34254373/204396828-47773c2b-5883-4601-878f-41e6ed24196f.png)
this is error

LovecraftianHorror:
Poked into this for a bit. It looks like it's failing to setup the clipboard provider since Wayland takes some special care to be used in `copypasta`

I can work on resolving this is a week or so

steven-omaha:
Still happening on latest `HEAD`, but with a different error message.

```
$ RUST_BACKTRACE=full cargo run --profile release-debug -- README.md 

     Running `target/release-debug/inlyne README.md`
thread 'main' panicked at src/main.rs:337:26:
called `Result::unwrap()` on an `Err` value: Renderer failed to redraw the screen

Caused by:
    0: Failed to acquire next swap chain texture
    1: The underlying surface has changed, and therefore the swap chain must be updated
stack backtrace:
   0:     0x5641f19957ac - std::backtrace_rs::backtrace::libunwind::trace::ha637c64ce894333a
                               at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/std/src/../../backtrace/src/backtrace/libunwind.rs:104:5
   1:     0x5641f19957ac - std::backtrace_rs::backtrace::trace_unsynchronized::h47f62dea28e0c88d
                               at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x5641f19957ac - std::sys_common::backtrace::_print_fmt::h9eef0abe20ede486
                               at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/std/src/sys_common/backtrace.rs:67:5
   3:     0x5641f19957ac - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::hed7f999df88cc644
                               at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/std/src/sys_common/backtrace.rs:44:22
   4:     0x5641f19c7e50 - core::fmt::rt::Argument::fmt::h1539a9308b8d058d
[...]

PR #130, opened 2023-05-27
Fix for Wayland, Make X11 and Wayland features

ripe-mangos:
- Fix for #55
 - Adds `x11` and `wayland` (default) features-- so users can opt out if they so choose

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,11 @@ repository = "https://github.com/trimental/inlyne"
 homepage = "https://github.com/trimental/inlyne"
 rust-version = "1.65"
 
+[features]
+default = ["x11", "wayland"]
+x11 = ["winit/x11", "copypasta/x11"]
+wayland = ["winit/wayland", "copypasta/wayland"]
+
 [dependencies]
 winit = { version = "0.28.5", default-features = false }
 wgpu = "0.16"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ homepage = "https://github.com/trimental/inlyne"
 rust-version = "1.65"
 
 [features]
-default = ["x11", "wayland"]
+default = ["x11"]
 x11 = ["winit/x11", "copypasta/x11"]
 wayland = ["winit/wayland", "copypasta/wayland"]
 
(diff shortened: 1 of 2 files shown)


CosmicHorrorDev:
Thanks for the PR! Were you able to reproduce the issue in #55 to confirm that this fixes it?

CosmicHorrorDev (changes requested):
Just got to take a look at the code. This is currently fixing things by removing clipboard support from all platforms which isn't desirable

`alacritty` should be a good example of how to handle this. From what I remember the event loop on wayland provides some clipboard context pointer that is used for enabling clipboard support since wayland is more fine-grained on access to these kinds of things

I've been meaning to setup a test system to reproduce the issue, so that I can verify that a fix works, but all the wayland systems I have setup have various compatibility layers setup to fix this kind of stuff. I'll work on setting up a more barebones system to reproduce tomorrow

trimental:
@CosmicHorrorDev the actual fix for this lies in this [line](https://github.com/alacritty/alacritty/blob/32ea98deeba5f0eb6c51062cfb0fd2caa802054d/alacritty/src/clipboard.rs#L34). 

Basically clipboard events need to be driven by the Wayland event loop but because winit doesn't allow you to do this, smithay clipboard creates a new event loop to drive this in a very hacky way but you have to provide a pointer to the Wayland display.

Would have thought I would have  remembered this since I'm the one who made smithay-clipboard but 🤷.

ripe-mangos:
@CosmicHorrorDev
 > Were you able to reproduce the issue in https://github.com/trimental/inlyne/issues/55 to confirm that this fixes it?

Yes, I'm using Hyprland 0.24.1 on Gentoo.

I don't know much about Wayland internals. This is just a quick, temporary fix until somebody that does can work on it.

[`4ed864b`](https://github.com/trimental/inlyne/pull/130/commits/4ed864b039851453637ee571f7d8f663e483b155) makes `x11` the default and `wayland` behind a feature wall.
As Wayland is still enthusiasts, I think that should be acceptable.

CosmicHorrorDev (approved):
This seems fine as a temporary workaround

Thanks, and congratz on your first PR here! :tada:

PR #132, opened 2023-05-28
Interpreter testing

CosmicHorrorDev:
Here is the aforementioned interpreter testing PR

All the changes are split out into individual commits to keep reviewing a bit easier as needed. The changes are basically

1. Shifting the layout (no code changes)
1. Splitting out nasty external state behind a separate trait, so that we can do dependency injection
2. Some boilerplate for being able to get the results from interpreting a single markdown document since the interpreter is meant to run continuously in a background thread
3. Slimming down a lot of noise in the debug impls for some things to make snapshot tests easier to review
    - The amount of noise reduced in this snapshot file diff is pretty telling (less than a third of the lines) https://github.com/trimental/inlyne/commit/2e4c5f5dade557fe9758c5b922604fcef8f2ca7e#diff-9baa87c0499d921080a8c793aed88c9fc5f15d7b1e5fe2e612f250deac1593e2
    - It should still be cheap to call these debug impls since I avoided any complex logic and extra allocations

I think interpreter tests are a pretty good candidate for snapshot testing since eyeballing the results of the output is a pretty good way to see that we're generating elements right, and they're very easy to update. Here's an example of the review screen if you haven't used `insta` before (it does mean using [`cargo-insta`](https://crates.io/crates/cargo-insta) for reviews though when snapshots do change. You don't _have_ to, but it's the easiest ways to handle it)

insta_review

You also don't have to use snapshot tests either since tests can just use the `interpret_md()` helper function

_(Side note: I only noticed from looking at the snapshot file that having a textbox with multiple anchors doesn't work, so the textbox with multiple footnotes only has the second anchor attached)_


CosmicHorrorDev:
Forgot to mention: we'll want to add custom debug impls for some of the other elements, but I was saving that for whenever we end up using other elements. `Image` probably has a massive `Debug` output since it holds the compressed image data

trimental:
Everything looks good. Thanks for the commit separation, made reviewing a lot easier. Insta seems like a very good choice for testing this.

 Slightly unrelated to this PR but does the repology svg render for you? Looks like it might be a windows specific ssl issue but the request for it is failing for me.

CosmicHorrorDev:
> Slightly unrelated to this PR but does the repology svg render for you? Looks like it might be a windows specific ssl issue but the request for it is failing for me.

Works fine for me on linux. I'm not sure what windows normally uses for ssl, but we could always switch to `rustls` instead. It should just a `default-features = false, features = ["rustls-tls"]` to switch

PR #133, opened 2023-05-28
Checklists shouldn't have a text prefix

CosmicHorrorDev:
> @CosmicHorrorDev **:** The first commit is just some twiddling to make `HtmlInterpreter` testable since it's a big bundle of mutable state, **and I don't want to accidentally regress something**

Whelp, this is certainly fitting. I noticed that #128 regressed checklists because _apparently_ lists not always getting a prefix based on the element stack was a ~~bug~~feature (whether that was intentional or not). This fix just makes it, so that processing an `"input"` checkbox will pop the pending list prefix since checklists get a custom prefix

This PR is based off of #132, so the only new commits are

- https://github.com/trimental/inlyne/commit/8154f9bdf8bfc57025b7b4d73b130d66894d79f6 - Adds a snapshot test to demonstrate that checklists incorrectly get a bullet prefix
- https://github.com/trimental/inlyne/commit/2770707e45c6443b6c603d679c01802cb8e15813 - Fixes the bug and updates the snapshot test indicating that the bulleted prefix is gone

At least now we have a test to catch any potential future regressions


trimental:
Ooo nice spot, thanks for getting this fixed. Really liking the Insta tests for the intepreter.

PR #136, opened 2023-05-31
Table Alignment

trimental:
Adds table item horizontal alignment as shown below:

image

Looks like wrapping is still occurring in some weird places, might just be some f32 nonsense with the bounds changing slightly between rendering and layingout

### DIFF
--- a/src/interpreter/mod.rs
+++ b/src/interpreter/mod.rs
@@ -327,8 +327,39 @@ impl TokenSink for HtmlInterpreter {
                             self.current_textbox
                                 .set_quote_block(Some(self.state.text_options.block_quote));
                         }
-                        "th" => self.state.text_options.bold += 1,
-                        "td" => {}
+                        "th" => {
+                            self.state.text_options.bold += 1;
+                            let align = tag
+                                .attrs
+                                .iter()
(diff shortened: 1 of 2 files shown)


CosmicHorrorDev (approved):
Looks great to me! I never even noticed that tables didn't handle alignment before

src/interpreter/mod.rs
```diff
@@ -327,8 +327,39 @@ impl TokenSink for HtmlInterpreter {
...
+                                .attrs
+                                .iter()
+                                .find(|attr| attr.name.local == local_name!("align"))
+                                .map(|attr| match attr.value.to_string().as_str() {
+                                    "left" => Some(Align::Left),
+                                    "center" => Some(Align::Center),
+                                    "right" => Some(Align::Right),
+                                    _ => None,
+                                })
+                                .unwrap_or_default();
```
nicoburns, line 342: I see this bit is repeated. Maybe it would make sense to make it a method on `tag` or `tag.attrs`?

CosmicHorrorDev: I was planning on refactoring things in a followup PR to use an `AttrIter` iterator since there's a lot of near duplicated code for attribute parsing scattered around the interpreter

PR #137, opened 2023-06-01
Remove `code-block-color`

CosmicHorrorDev:
This is just a vestigial configuration option from back before we used `syntect` to handle code highlighting. It's only use was for code blocks which are now always handled by `syntect` instead


trimental:
This was kept just in case the background color wasn't set in a theme but i haven't seen any examples of this. And if a theme doesn't set it, we should probably just respect that anyway. Looks good 👍.

PR #139, opened 2023-06-03
Refactor with `AttrIter`

CosmicHorrorDev:
Refactors to move the bulk of attribute and style matching outside of the interpreter. This adds in an `AttrIter` and `StyleIter` (along with a couple of helper functions) to ease handling attributes and styles

Note: This removes handling `text-align` as an attribute since it's a CSS style that isn't a valid attribute


trimental:
Why is text-align not a valid attribute?

trimental:
Maybe in the future we want to re-add text-align but for now it's probably best not to scatter style parsings for text-align aimlessly. I'm currently working on nested tables support which has made some changes to the interpreter. Once I'm done I'll merge these together. Everything looks great.

CosmicHorrorDev:
The `text-align` stuff wasn't working correctly before though. It was handling it as an attribute when it's not an attribute. E.g. the interpreter on current `main` will handle

```html
<p text-align="center">Text</p>
```

as centered text when a browser would ignore the attribute and use default alignment instead. We can properly handle it if people are already using it in the wild, but it needs to be moved to style parsing (which would be easy enough to handle)

Sounds good for merging after your pending work though :+1:

CosmicHorrorDev:
Hiya @trimental! I'm guessing you''re probably just busy with life stuff at the moment. Are you fine if this gets merged before the table changes? This is the only larger refactor on the interpreter I have planned, and I want to start working on some smaller interpreter related changes that would conflict with this

CosmicHorrorDev:
This is blocking some other work that I want to get done, so I'm going to go ahead and merge this, then I'll be working on trying to revive those table changes you had before (and then updating dependencies, fixing a couple of lingering issues, and publishing a release??? :smiling_imp:)

Anyhoo I'm assuming you're just busy with life, so hope everything is going well! I'll try to keep the lights on around here till you're back

PR #143, opened 2023-06-15
Improve error message when no file path is provided

abclop99:
Makes the argument required by `clap` so that the program prints a useful error message and exits when the `file` argument is missing.

Before:

```shell
$ inlyne
thread 'main' panicked at 'required', src/opts/cli.rs:101:58
```

After:

```shell
$ inlyne
error: the following required arguments were not provided:
  <FILE>

Usage: inlyne <FILE>

For more information, try '--help'.
```


CosmicHorrorDev (approved):
Looks great. Thanks!

PR #147, opened 2023-06-18
More robust file watching

CosmicHorrorDev:
Before file watching would keep watching a file even after it was renamed. File watching would also fail permanently if you did something like removing the original file and replacing with a new file after a short delay

These changes handle all of the above cases by `.unwratch()`ing the file after name changes (the path associated with the watcher is still the original path name even after a rename) and looping continuously to re-register the watcher after the original file is renamed/removed

I also added a reminder to fix watching the file after changing the file by clicking a markdown link from within the current file, but haven't handled that quite yet

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1741,6 +1741,7 @@ dependencies = [
  "dark-light",
  "dirs 5.0.1",
  "env_logger",
+ "filetime",
  "fxhash",
  "glyphon",
  "html-escape",
@@ -1762,6 +1763,7 @@ dependencies = [
  "serde_yaml",
  "syntect",
  "taffy",
+ "tempfile",
  "tiny-skia",
  "toml",
  "twox-hash",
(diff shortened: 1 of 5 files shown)


CosmicHorrorDev:
> I also added a reminder to fix watching the file after changing the file by clicking a markdown link from within the current file, but haven't handled that quite yet

Figured I would go ahead and get around to this. :crossed_fingers: that the test passes CI

CosmicHorrorDev:
Re-ran CI 5 times which means that the test suite passed 15 times in a row across three platforms, but it seems like the new test may be naturally flaky unfortunately. If it becomes a problem in the future then I can just wrap the whole thing in a set amount of retries to drastically reduce the failure rate

Also it looks like the x86 mac runner is the only runner that rebuilds everything on every run even though it has a cache hit, so that's odd

CosmicHorrorDev:
I'm confident that the test should be reasonable reliable now. If there are any issues in the future then I can investigate and may wrap the test in a fixed amount of retries

PR #150, opened 2023-07-04
Improve syntax highlighting

CosmicHorrorDev:
This fixes a couple of issues that both conveniently manifest in `reqwest`'s README

1. The default syntax definitions are missing a lot of common languages (including TOML)
2. Fenced code blocks may have an info string like ```` ```rust,ignore```` since some projects use the README for doc stuff. `comrak` doesn't recognized this as a Rust block although GitHub does

## `main`

image

## This PR

image

Yes the Rust highlighting is this basic for this example. It turns out that syntect is stuck using several year old syntax definitions for a lot of languages since Sublime Text has made changes to syntax definitions that would require significant changes to syntect

Part of this is also from the default dark theme. `two-face` is now getting pulled in which includes a lot more themes though, so I'll add those in a future PR

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1765,6 +1765,7 @@ dependencies = [
  "tempfile",
  "tiny-skia",
  "toml",
+ "two-face",
  "twox-hash",
  "usvg",
  "wgpu",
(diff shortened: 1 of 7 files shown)


CosmicHorrorDev:
Gonna go ahead and slap the merge button for this one

There's still some followup work that I want to get through. Namely:

- Adding a `syntect-no-defaults` feature flag to `comrak` that exposes the syntect adapter without pulling in syntect's default embedded assets (because we're not using them anymore)
- Making use of `two-face`'s extra themes as well since people would probably appreciate the extra customization

I don't think either of those should block merging though since this PR is an obvious improvement in terms of syntax highlighting

PR #167, opened 2023-11-01
Update `wgpu` to the latest version

CosmicHorrorDev:
Bumps `wgpu` and friends to the latest version. This is just a sanity check to see if this resolves any `wgpu` related bugs


CosmicHorrorDev:
Looks like this will be hard to manage before updating everything else

PR #195, opened 2023-12-13
Late night refactors

CosmicHorrorDev:
More refactorings and cleanups. Some somewhat more expensive things should be more responsive and the test suite now finishes in 0.26 seconds for me :rocket::rocket:


CosmicHorrorDev:
Whelp looks like those delays will still need to be increased back

Issue #188, opened 2023-12-10
`extra` keybindings should have precedence over `base`


CosmicHorrorDev:
The purpose of `extra` was to give the user control over changing just a couple of keybindings here and there while still making use of the defaults (`base`). The current design partially fulfills this purpose by giving the option to add additional keybindings, but it's limited in that you can't override individual keybindings. It would be nice if `extra` keybindings had a higher precedence than `base` in that keybindings that appear in both will use the `extra` version instead of choking on a conflict

Notably this won't address conflicts where a keybinding in one _prefixes_ some keybinding in the other, but I don't know of a good resolution there. I think as it stands now any additional keybindings will have to be added with a semver breaking bump

Issue #162, opened 2023-09-19
Feature: a way to go back


sidju:
Since you can click links to open other local files a way to go back afterwards would be good.

CosmicHorrorDev:
This should be possible now in `main` with the `Alt+<left or right arrow>` or the `bp` or `bn` keybinds

Issue #131, opened 2023-05-27
Add an `--edit-config` flag


CosmicHorrorDev:
(Opening an issue for this, so that I don't forget about it)

It would be cool if:

- The [config] file was automatically created (no copying neccessary)
- And ideally if there was a keyboard shortcut or some other way to open to file (or at least view it's path)

_Originally posted by @nicoburns in https://github.com/trimental/inlyne/issues/124#issuecomment-1556263248_

kokoISnoTarget:
Should the `--edit-config` flag circumvent opening a markdown file? e.g. `inlyne --edit-config` vs `inlyne some.md --edit-config`.

CosmicHorrorDev:
Yes that would make sense, but I'm actually leaning towards making it a `config` subcommand instead, so that we can add other actions in the future like `config set <key> <val>` or `config path`, so I think `--edit-config` would fit best as `inlyne config open` probably?

I think overall I'm thinking

```
inlyne [FILE] [COMMAND]

Commands:
config  do config things
view    view a markdown document

Arguments:
[FILE]  markdown file to view
```

That way you can do things like

``` bash
$ inlyne config open  # opens the config in an editor
$ inlyne README.md    # views the file "README.md"
$ inlyne view config  # views the file "config"
```

It does add in the edge cases of passing a file and a command, and passing neither which should both be CLI errors

Also worth noting that we can switch back to `clap`'s derive API since we no longer need the full flexibility of the builder API

Issue #53, opened 2022-10-19
[Suggestion] History and Navigation


nerdachse:
Hey there!

## Motivation

After #50 is closed, clicking a link to a local `md` file will open it in the current window.

I think it makes sense to add some lightweight "history" and "navigation" in order to be able to go back to the previously opened file.

## Other alternatives

Well, if every markdown file has a link to the previous one, this is a none issue. Like if you create a strict hierarchy. But it relies on the content of the markdown files, which is outside our control.

I think it makes sense to add this, especially now (or rather soonish ;)) that the content of the current window is overwritten.

## To decide

Shortcuts for navigation and maybe even some kind of menu-bar/ui buttons?

trimental:
I think navigation shortcuts would be a good idea however I'm unsure whether ui buttons would be a good fit. One of the reasons (other than being lazy) of avoiding creating a ui interaction bar or something similar, is that it would take up screen space and add visual complexity. 

I like the idea of inlyne being a viewport to the markdown page and nothing more. More of a CLI app that spawns a window than a GUI app that's launched from the terminal.

History shouldn't be hard to track and the framework to implement customisable shortcuts was already made by @LovecraftianHorror. Overall this should be another easy feature to implement :)

CosmicHorrorDev:
This should be possible now in `main` with the `Alt+<left or right arrow>` or the `bp` or `bn` keybinds

PR #200, opened 2023-12-20
fix: filter file watcher duplicate events

### LINKED ISSUE #198: Live render partially working
# Description

Sometimes when saving a file, the live reload will not render the whole file (screen 3) or just a blank screen (screen 2).
The initial render - when starting the app - is always ok (screen 1).

Triggering the reload again will end up in one of the 3 mentioned cases.
I noticed saving with modifications will more often lead to a correct render than saving without.

It seems that the larger the file the easier it is to reproduce. While testing, I started to encounter the issue with ~250 lines, but it was very rare. 

I'd like to help more but couldn't find any log of why that happens, neither stdout/stderr or files.

From an outside point of view it seems to me like there is some non-determinism going on which causes inlyne to sort of exit early. 

# Additional details

<details><summary>Screenshots</summary>
<p>

Screen 1, working correctly on open
image

Screen 2, blank render on live reload
image

Screen 3, partially rendered after live reload (notice the size of the scrollbar compared to screen 1)
image

</p>
</details> 

I run ubuntu with i3wm and neovim v0.9.4 (if that's any relevant).

```sh
$ uname -a
Linux xubuntu 5.15.0-91-generic #101-Ubuntu SMP Tue Nov 14 13:30:08 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
```
```sh
$ inlyne --version
inlyne 0.3.3
```
Also happens on main, 0.3.2

I mainly tested against the [ratatui breaking change file](https://github.com/ratatui-org/ratatui/blob/main/BREAKING-CHANGES.md) and the other md files in the repo.

---

Please tell me if I can be of any more help!

---

EDIT: on version 0.2.1, I don't have any blank screen, sometimes a partial render (but very rare). Hope that helps.

Valentin271:
This fixes #198

As discussed in the issue, I refactored the file watcher to use [notify_debouncer_full](https://docs.rs/notify-debouncer-full/latest/notify_debouncer_full/index.html).

I did so with as few modifications as possible to avoid breaking anything. The most interesting part is the `DebounceEventHandler` impl in which I basically filtered unwanted and duplicate events.
The debouncer timeout of 10ms seems good for up to 100 ~ 150Kb files on my machine.

All unit tests passed locally. I've also tested with the files I was having issues with.

I also added logs I found interesting for debugging and changed the log format to show milliseconds since the file_watcher is always faster than a second.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1161,6 +1161,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "file-id"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6584280525fb2059cba3db2c04abf947a1a29a45ddae89f3870f8281704fafc9"
+dependencies = [
+ "windows-sys 0.48.0",
+]
+
+[[package]]
 name = "filetime"
 version = "0.2.22"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1878,6 +1887,7 @@ dependencies = [
  "lyon",
  "lz4_flex",
  "notify",
+ "notify-debouncer-full",
  "open",
  "pollster",
  "pretty_assertions",
@@ -2499,6 +2509,19 @@ dependencies = [
  "mio",
  "walkdir",
  "windows-sys 0.48.0",
+]
+
+[[package]]
+name = "notify-debouncer-full"
+version = "0.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "49f5dab59c348b9b50cf7f261960a20e389feb2713636399cd9082cd4b536154"
+dependencies = [
+ "file-id",
+ "log",
+ "notify",
+ "parking_lot",
+ "walkdir",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -67,6 +67,7 @@ two-face = "0.3.0"
 # Required for WGPU to work properly with Vulkan
 fontdb = { version = "0.13.1", features = ["fontconfig"] }
 human-panic = "1.2.2"
+notify-debouncer-full = { version = "0.3.1", default-features = false }
 
 [dependencies.glyphon]
 version = "0.2"
--- a/src/file_watcher/mod.rs
+++ b/src/file_watcher/mod.rs
@@ -50,7 +50,7 @@ impl DebounceEventHandler for MsgHandler {
                 // Remove is more interesting than Modify
                 for ev in events {
                     match ev.event.kind {
-                        EventKind::Modify(_) => {
+                        EventKind::Create(_) | EventKind::Modify(_) => {
                             let _ = selected_event.get_or_insert(ev.event);
                         }
                         EventKind::Remove(_) => {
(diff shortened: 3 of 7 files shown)


CosmicHorrorDev:
Seems to be failing on mac, but all the file watching logic is notorious for platform specific quirks. I'll go ahead and re-run the failed jobs, but I doubt they'll pass this time (more so for my sanity). I won't be able to take too close of a look till after holidays, but I'll still be around to hit the CI button or for reviews (I also think CI should still run in your fork when you push to it for whenever I'm not around)

Valentin271:
Alright no problem. I had a feeling it would only work on my machine, it's good you have a CI to check that. I'll fix them.

It does run on my fork after enabling it.

CosmicHorrorDev:
Great! And do note that the watcher tests are a bit weird. They're flaky when using short delays, but longer delays suck for dev-feedback loops, so it starts with shorter delays and retries the tests a few times with increased delays on failure

Just in case the logs for the failed tests look confusing

Valentin271:
Alright good to know, I'm currently looking at the tests and I'm suspecting the delay is too short.

CosmicHorrorDev:
Found this issue thread. Looks like it could be relevant

https://github.com/notify-rs/notify/issues/150#issuecomment-495930414

CosmicHorrorDev:
I just pushed a commit that should fix the `sanity` file watching tests since it looks like we were ignoring `Create` events and that was the only one getting sent

Aside: trace logs are currently getting filtered out from tests, but I'm starting to think that it'd be a good idea to retain them still

CosmicHorrorDev:
Eh went ahead and changed that too. We really need all the info we can get when it comes to CI-driven development :upside_down_face:

CosmicHorrorDev (approved):
CI is green. Hope you didn't mind all the commits :partying_face: 

I'm happy to merge this as-is. Most of the issues seem to stem from MacOS sending create events instead of modify events when using the debouncer from some reason

Valentin271:
> We really need all the info we can get when it comes to CI-driven development

Yep, I agree.

> Hope you didn't mind all the commits

No problem go ahead.

---

> Most of the issues seem to stem from MacOS sending create events instead of modify

From notify_debouncer_full docs:
> Doesn’t emit Modify events after a Create event

> I'm happy to merge this as-is

Can confirm it still works, happy to see this merged then :)

CosmicHorrorDev:
I want to offer a huge thanks for all your help on this :)

Valentin271:
You're welcome, I'm glad I could help. This is a great tool that I really love and want to help with and see grow (or at least not abandoned).

CosmicHorrorDev:
There are plenty of things to work on to say the least. I can tag you in some issues that may be easier to start out with

Valentin271:
I've seen a few already, but I wouldn't mind if you tagged the good first issues.

CosmicHorrorDev:
Keep in mind that features don't have to be perfect to open a PR. I'm more than happy to help with draft PRs. From looking over the issues I see:

- Some keybindings stuff if you're comfortable with recursion
    - #188
- Adding history to our navigation
    - #162
    - #53
- Adding an `--edit-config` flag
    - #131
- Adding double/triple click although that may be complicated
    - #43

I picked those because they're either well tested or generally well isolated

I also have a lot of other things in an obsidian notebook, so I can start opening issues for those over the next few days. Feel free to reach out in whichever issues you want to take on so that we can talk about the general design/approach

CosmicHorrorDev:
Opened some new issues that include a lot of potentially good starter issues along with actually labeling easy(ish) issues

https://github.com/trimental/inlyne/issues?q=is%3Aopen+is%3Aissue+label%3AE-easy

PR #209, opened 2024-01-07
refactor: migrate from `log` to `tracing`

### LINKED ISSUE #206: Switch from `log` to `tracing`
`tracing` generally has many more features and a much more vibrant ecosystem that I want to take advantage of. The initial switch can just more-or-less mimic how our current logs look, but in the future we can also tee logs off to files and track spans and all

Valentin271:
Fixes #206

I've migrated `log` to `tracing`.

This is just a simple replacement for now, the logging functions and looks the same. `INLYNE_LOG` still controls the logging and the syntax is the same.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1057,19 +1057,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "env_logger"
-version = "0.10.0"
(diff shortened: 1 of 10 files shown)


CosmicHorrorDev (approved):
Looks great. Thanks!

PR #210, opened 2024-01-08
chore: update taffy

Valentin271:
Updates taffy to the latest crates.io version.

There is no breaking change summary so I had to look in taffy's internal code to figure out which feature went were.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3990,8 +3990,9 @@ dependencies = [
 
 [[package]]
 name = "taffy"
-version = "0.3.11"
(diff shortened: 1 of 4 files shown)


CosmicHorrorDev:
Thanks for this! I can take a look after I get off work today

CosmicHorrorDev (approved):
Everything seems to be working. Thanks!

The glyphon update will likely be harder since I think there are a few other dependencies that share `wgpu` as a dependency that may have to be updated together

Valentin271:
Took a look at it before pushing this PR (I originally wanted to update both in the same PR), it is going to be harder, there are more compiler errors right upfront. expect a PR _maybe_ next week.

nicoburns:
Worth noting that this updates to an *older* Taffy version rather than a newer one. Which is perfectly reasonable if you want to use the released version. But that's why there's no breaking change notes: the changes haven't been released yet!

PR #212, opened 2024-01-09
fix: fix default log level back to info

Valentin271:
Somehow, before opening #209, I forgot to put the default log level back to what is was, that is INFO.


CosmicHorrorDev (approved):
Thanks!

PR #213, opened 2024-01-09
fix: reset scroll on markdown navigation

### LINKED ISSUE #202: Scroll doesn't get reset when navigating
When navigating different markdown files the scroll offset of the original file is retained which is makes for a pretty confusing transition. Once history is properly added it would probably make sense to retain each session's scroll separately in some list of history entries

## Demo

https://github.com/trimental/inlyne/assets/30302768/c3e3e99a-72bd-4f8c-9b2b-629cf53964d9

Valentin271:
Fixes #202

Reset the scroll to top when navigating markdown files.

I think it's worth adding a test for this but I don't know where that would be.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -521,6 +521,9 @@ impl Inlyne {
                                                         &self.opts.file_path,
                                                         contents,
                                                     );
+                                                    // TODO: Once and if history is implemented,
+                                                    // old scroll_y might be stored there
+                                                    self.renderer.set_scroll_y(0.);
                                                 }
                                                 Err(err) => {
                                                     tracing::warn!(


src/main.rs
```diff
@@ -521,6 +521,9 @@ impl Inlyne {
                                                         &self.opts.file_path,
                                                         contents,
                                                     );
+                                                    // TODO: Once and if history is implemented,
+                                                    // old scroll_y might be stored there
+                                                    self.renderer.set_scroll_y(0.);
                                                 }
```
Valentin271, line 527: Note I didn't call `window.request_redraw()` as in `scroll_pixels`. It worked without it for me.
Furthermore, when adding the call it blinked, there is a few frames where I can see the original md at the top and then it switches.

My guess is the window is redrawn by the above call `watcher.update_file` once md is interpreted.

CosmicHorrorDev (approved):
Looks great, thanks!

It would be nice to add a test, but this part of the code doesn't really have any testing infrastructure setup at the moment. I've been slowly working on getting things more testable, but only bits and pieces support easy testing for now

PR #214, opened 2024-01-20
chore: update glyphon to 0.3

Valentin271:
This updates glyphon the latest version.

There is still one change I need help with (marked below).

EDIT: I'll test the changes locally once I can get it to compile.

### DIFF
--- a/src/text.rs
+++ b/src/text.rs
@@ -9,7 +9,7 @@ use crate::utils::{Align, Line, Point, Rect, Selection, Size};
 
 use fxhash::{FxHashMap, FxHashSet};
 use glyphon::{
-    Affinity, Attrs, AttrsList, BufferLine, Color, Cursor, FamilyOwned, FontSystem, Style,
+    Affinity, Attrs, AttrsList, BufferLine, Color, Cursor, FamilyOwned, FontSystem, Shaping, Style,
     SwashCache, TextArea, TextBounds, Weight,
 };
 use smart_debug::SmartDebug;
@@ -93,8 +93,8 @@ impl Default for TextBox {
 
 pub struct CachedTextArea {
     key: KeyHash,
-    left: i32,
-    top: i32,
+    left: f32,
+    top: f32,
     bounds: TextBounds,
     default_color: Color,
 }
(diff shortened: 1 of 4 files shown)


Valentin271:
This is the part where I need help.

https://github.com/trimental/inlyne/pull/214/files#diff-32520862cbc569bbe7bbe69197fb360bdf01d9eabd5839d24db034724fa3e766R774-R776

The main problem is that `grow` is private, I'm pretty sure this was the change made in trimental's fork.
If so we won't be able to update glyphon right?

CosmicHorrorDev:
I believe that you should be able to remove us calling `.grow()` since growing should be handled internally now. The relevant PR adding the feature upstream is

https://github.com/grovesNL/glyphon/pull/34

Valentin271:
You're right, this seems to be working locally without grow.

Thanks for the PR link I'll check it out!

src/text.rs
```diff
@@ -651,7 +652,7 @@ impl TextCache {
                             .metadata(section.index),
                     )
                 }
-                let buffer_line = BufferLine::new(line_str, attrs_list);
+                let buffer_line = BufferLine::new(line_str, attrs_list, Shaping::Basic);
```
CosmicHorrorDev, line 655:
```suggestion
                let buffer_line = BufferLine::new(line_str, attrs_list, Shaping::Advanced);
```

It looks like advanced shaping is what includes font fallback, so we'll want that (an example to check is the :eyes: emoji in the `README.md` since emojis are usually only included in a handful of different fonts)

CosmicHorrorDev (approved):
Everything looks good after that one comment is resolved

Huge thanks for getting through these. It's nice to be able to be able to publish to crates.io again

src/text.rs
Valentin271: I wasn't sure about this one. It renders correctly with `Basic` for me. I guess that means my system font supports emojis...

Valentin271:
You're welcome! I wanted to fix that before adding anything else that might be broken by updates.

src/text.rs
CosmicHorrorDev: Ah, it doesn't for me, so easier for me to spot 😅 (Eventually I'd like to get testing setup where we can render the output to something like a PNG and use an embedded set of fonts so that we can actually programmatically check these things)

Valentin271: That's be nice yeah. Even if it means using a shell script or some external tool that'd be better than nothing.

PR #216, opened 2024-01-21
chore: refactor CI runs

### LINKED ISSUE #203: Restructure CI runs
Currently CI runs will run an identical validation check over all the supported platform+toolchain combos, but we don't really get anything from checking formatting and clippy on all of those. Instead it would be better if we had those two jobs split out and run on ubuntu together (since formatting is a near instant check)

Valentin271:
Closes #203 

- test
  - cargo test
  
- lint
  - typos
  - fmt
  - clippy
  
- build-macos-arm (now builds on `stable` and `beta`)
  - only build

Note that macos-arm runner exist since october 2023 but are paid and in beta for now: https://github.com/orgs/community/discussions/69211.
Once and if they get released publicly, macos-arm can be merged with build.

See https://github.com/Valentin271/inlyne/actions/runs/7600723247 for the latest successful run from my tests.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -4,10 +4,10 @@ on: [push, pull_request]
 
 env:
   CARGO_TERM_COLOR: always
-  RUSTFLAGS: "-C debuginfo=0"
+  RUSTFLAGS: '-C debuginfo=0'
 
 jobs:
-  build:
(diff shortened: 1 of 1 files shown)


CosmicHorrorDev (approved):
Looks great. Thanks as always!

Issue #197, opened 2023-12-14
thread `main` panicked at src/text.rs:206:17 attempt to subtract with overflow


CosmicHorrorDev:
This is from the unreleased text changes, so this bug isn't out yet. It would be good to fix it before it even gets released. Expanding and scrolling through collapsible sections usually does the trick. Needs a dev build to actually fail on the overflow since release builds disable that check by default

```console
$ RUST_BACKTRACE=1 c r -- ../typst/README.md
    Finished dev [unoptimized + debuginfo] target(s) in 0.14s
     Running `target/debug/inlyne ../typst/README.md`
thread 'main' panicked at src/text.rs:206:17:
attempt to subtract with overflow
stack backtrace:
   0: rust_begin_unwind
             at /rustc/1a06ac5b5d7c9331e8de1aa1fd7e9d3533034b44/library/std/src/panicking.rs:645:5
   1: core::panicking::panic_fmt
             at /rustc/1a06ac5b5d7c9331e8de1aa1fd7e9d3533034b44/library/core/src/panicking.rs:72:14
   2: core::panicking::panic
             at /rustc/1a06ac5b5d7c9331e8de1aa1fd7e9d3533034b44/library/core/src/panicking.rs:127:5
   3: inlyne::text::TextBox::find_hoverable
   4: inlyne::Inlyne::find_hoverable::{{closure}}
   5: core::option::Option<T>::and_then
   6: inlyne::Inlyne::find_hoverable
   7: inlyne::Inlyne::find_hoverable::{{closure}}
   8: core::option::Option<T>::and_then
   9: inlyne::Inlyne::find_hoverable
  10: inlyne::Inlyne::run::{{closure}}
  11: winit::platform_impl::platform::x11::EventLoop<T>::drain_events::{{closure}}::{{closure}}
  12: winit::platform_impl::platform::sticky_exit_callback
  13: winit::platform_impl::platform::x11::EventLoop<T>::drain_events::{{closure}}
  14: winit::platform_impl::platform::x11::event_processor::EventProcessor<T>::process_event
  15: winit::platform_impl::platform::x11::EventLoop<T>::drain_events
  16: winit::platform_impl::platform::x11::EventLoop<T>::run_return::single_iteration
  17: winit::platform_impl::platform::x11::EventLoop<T>::run_return
  18: winit::platform_impl::platform::x11::EventLoop<T>::run
  19: winit::platform_impl::platform::EventLoop<T>::run
[...]

PR #217, opened 2024-01-21
fix(text): fix substract with overflow panic

Valentin271:
Fixes https://github.com/trimental/inlyne/issues/197

This change avoids the panic, _visually_ the behavior is the same.
However I'm not sure this is correct, because it means sometimes it won't be subtracted (and we get a wrong index?).

---

I don't get why we would need to subtract one to the index of the glyph under the cursor only if the cursor is associated with the previous frame. Also removing this `if Affinity::Before` altogether doesn't seem to change anything ...
Have I misunderstood something in my reasoning?

### DIFF
--- a/src/text.rs
+++ b/src/text.rs
@@ -177,6 +177,7 @@ impl TextBox {
         }
     }
 
+    /// Returns the [`Text`] in the given [`TextSystem`] with the cursor over it, if any.
     pub fn find_hoverable<'a>(
         &'a self,
         text_system: &mut TextSystem,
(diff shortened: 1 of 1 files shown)


CosmicHorrorDev (approved):
I believe this is the correct behavior. The affinity stuff is used to determine which run the cursor should be a part of when it's between two text runs. Subtracting one (kinda shaky logic that it's exactly one) ensures that the cursor falls on the previous text run when between two, and the only value that saturates would be zero which should only happen when you're already on the first text run

Valentin271:
Ok I think I'm misinterpreting what a text run is. With your explanation it seems more like a chunk of text (and now it makes sense to be between two elements, selecting the previous glyph) whereas I understood it as kind of a render of the document.

Issue #141, opened 2023-06-04
Underline alignment regression


CosmicHorrorDev:
cc @trimental

Just noticed that the PR to add alignment for tables regressed alignment for text vs its underline

**Repro**

```html
<h1 align="center">Some text</p>
```

**Rendered**

![image](https://github.com/trimental/inlyne/assets/30302768/1ce5c3ec-4494-48cf-a51a-529b651a4e62)

You can see that the text is center aligned while the underline is left aligned

This seems to be due to removing the alignment from the `Key` which was used to set the alignment of the `buffer_line`. Admittedly I'm not really familiar enough with the renderer to know how to fix things without likely regressing other parts

trimental:
Yep haha. Currently trying to fix this issue. Getting close to honing in on it but it's a tricky one. Going to need more time to sort it out.

PR #218, opened 2024-01-21
fix(table): table alignment wasn't reset properly

Valentin271:
While working on https://github.com/trimental/inlyne/issues/141 I found a bug in the table alignment that I would like to fix first.

Each cell, if no explicit alignment specified, retain the alignment of the previous cell. This also occurred across table.

---

Assume the following table:
```
| left default | left forced | centered | right | left default |
| ------------ | :---------- | :------: | ----: | ------------ |
| text         | text        |   text   |  text | text         |
```

On main:
image

With this fix:
image

### DIFF
--- a/src/interpreter/mod.rs
+++ b/src/interpreter/mod.rs
@@ -290,14 +290,12 @@ impl TokenSink for HtmlInterpreter {
                         }
                         "th" => {
                             self.state.text_options.bold += 1;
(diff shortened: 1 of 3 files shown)


CosmicHorrorDev (approved):
Was going to mention adding a regression test, but you're already way ahead of me. Thanks!

Issue #211, opened 2024-01-09
regression: no more diff coloration


Valentin271:
In inlyne 0.3.3, ` ```diff` were rendered as follow (the whole line colored).
![image](https://github.com/trimental/inlyne/assets/36198422/eda80ce5-1ec1-40d9-b85f-18f6774c7d32)

With inlyne main, ` ```diff` renders as follow (only `+` colored).

![image](https://github.com/trimental/inlyne/assets/36198422/806e792c-b328-4d24-a619-b8a2a1ec996c)

CosmicHorrorDev:
Just guessing from where things regressed, but it's probably just that the new default light theme doesn't support highlighting diffs correctly (which isn't surprising because the themes embedded with `syntect` are of mixed maintenance status)

We have access to [a lot more themes through `two-face`](https://docs.rs/two-face/latest/two_face/theme/fn.extra.html) now (thanks to `bat`), so it would probably be good to evaluate and switch over to (potentially a subset of) those themes

Valentin271:
Good to know, i'll take a look at it when possible.

I thought of the theme support too but the colored `+` threw me off. I'll check and see.

Valentin271:
I did a little bit of research on that today, I can confirm this simply comes from `InspiredGithub` theme not supporting diff correctly.

I think we should switch to using `two-faces` theme so people have more choice.

Do you remember why you switched to `InspiredGithub`? I think default dark and light `color_highlighter` should be variants of the same theme (e.g. `Base16Ocean{Dark,Light}`, `OneHalf{Dark,Light}`).

CosmicHorrorDev:
We originally switched away from a light/dark variant theme due to the light theme having very low-contrast colors (Base16Mocha IIRC)

I picked inspiredgithub since it seemed to have much better contrast, but that was before we added `two-face`. I've got some free time today, so I can work on integrating two-face's themes

Valentin271:
I thought `Base16OceanLight` had low contrast too.
I tried some themes today, I think `monokai` and `OneHalf` (but diff colors are too muted on this one imo) are the best options (that have variants, support diff and have decent colors overall).

Here is a quick comparison:
![image](https://github.com/trimental/inlyne/assets/36198422/4746e0cf-7ce9-4176-8fda-2ced766a6324)

CosmicHorrorDev:
#219 exposes a lot more themes if you want to take a look at those too

[...]

PR #223, opened 2024-01-25
feat: support bg color in code blocks

Valentin271:
In order to fix https://github.com/trimental/inlyne/issues/211, I've tried adding support for background colors.

I just can't get the final `Text` render to work because I'm not sure where it is.


CosmicHorrorDev:
I'm too busy with work to take a look today/tomorrow, but I should have time to try and get things working this weekend

The renderer is pretty foreign to me too, so I don't have much guidance without digging in myself unfortunately 😅

Valentin271:
No problem at all! Review when you have time and motivation :slightly_smiling_face:.

In the meantime I'll still try to get this working.

CosmicHorrorDev:
So, this is the area where the background of the entire codeblock gets drawn, so it'll also be where the background color for the individual lines get drawn as well

https://github.com/trimental/inlyne/blob/main/src/renderer.rs#L249-L274

I also think the default background color and style changes can be dropped in `text.rs` since that's handling different stuff

---

Feel free to ask questions if things are still unclear. I'd also be happy to take a stab at getting things working if you get stuck

Valentin271:
Thanks for the help, I've got some background rendering to work. However there is a lot of bug right now and the code is absolutely hideous.

- Doesn't work with soft wrap
- For some reason background color of normal code is dark red (I've checked this is not on syntect end's), but the logs shows `bg_color` gets set to this at some point
- The code block has a higher height the the sum of `Text`s, making it visible at the bottom of a diff
- I think the colored rectangle is not very well positioned on the line, i.e. It probably need to be a tiny bit lower

Here is a screenshot of the render as of now:
image

To do this I'm actually missing some position/size helpers on `Text`. It there any that I've missed?
Also I'd need to iterate over lines, I wouldn't be happy with letting `if text == "\n"` in a production code.

CosmicHorrorDev:
Thanks for all your work on this! I should have some time to look over things either today or tomorrow

CosmicHorrorDev:
Sorry for the delay. I'm on vacation, so I probably won't have much time to look into this for the next week or so

PR #224, opened 2024-01-26
feat: extra keybindings override base

Valentin271:
Fixes https://github.com/trimental/inlyne/issues/188

I suspect this PR will be easier to review by reviewing commits individually (that's why I left them) to understand the reasoning behind the changes. I also left some PR comments to explain some changes.

I refactored the code a bit to be easier to work with, hope that's okay with you.

### DIFF
--- a/src/keybindings/keybindings.rs
+++ b/src/keybindings/keybindings.rs
@@ -49,7 +49,14 @@ impl From<KeybindingsSection> for Keybindings {
                 base.0 = base
                     .clone()
                     .into_iter()
-                    .filter(|(_, combo)| !combo.starts_with(extra_combo))
+                    .filter(|(_, combo)| {
+                        if combo.starts_with(extra_combo) {
+                            tracing::debug!("Base keybinding {combo} ignored in favor of extra keybinding {extra_combo}");
+                            false
+                        } else {
+                            true
+                        }
+                    })
                     .collect();
             }
 
--- a/src/keybindings/keybindings.rs
+++ b/src/keybindings/keybindings.rs
@@ -10,6 +10,10 @@ impl Keybindings {
         Self(bindings)
     }
 
+    pub fn iter(&self) -> std::slice::Iter<'_, (Action, KeyCombo)> {
+        self.0.iter()
+    }
+
     #[cfg(test)]
     pub fn empty() -> Self {
         Self(Vec::new())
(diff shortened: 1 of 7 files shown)


src/keybindings/keybindings.rs
```diff
@@ -0,0 +1,121 @@
...
+        let mut base = value.base;
+
+        if let Some(extra) = value.extra {
+            for (_, extra_combo) in extra.iter() {
+                base.0 = base
+                    .clone()
+                    .into_iter()
+                    .filter(|(_, combo)| !combo.starts_with(extra_combo))
+                    .collect();
+            }
```
Valentin271, line 54: This is the relevant part of the PR, we eliminate `base` bindings that appear in `extra`

src/opts/config.rs
```diff
@@ -66,9 +66,10 @@ impl Default for LinesToScroll {
     }
 }
 
-#[derive(Deserialize, Debug, Default, PartialEq)]
+#[derive(Deserialize, Clone, Debug, Default, PartialEq)]
 pub struct KeybindingsSection {
-    pub base: Option<Keybindings>,
+    #[serde(default)]
+    pub base: Keybindings,
```
Valentin271, line 72: This should be the exact same as before but with added benefits:

- `base` is always something, no need to `unwrap` etc...
- if the section doesn't exist, defaults are loaded. This avoids using `unwrap_or_default` everywhere which is error prone imo and simpler to read.
- The app always has `base` keybindings so it makes sense semantically

doc: https://serde.rs/field-attrs.html#default

src/keybindings/keybindings.rs
CosmicHorrorDev: Probably good to log the combos that are getting filtered to make debugging easier

CosmicHorrorDev (approved):
Thanks! Really happy with all of this :)

In general refactors are AOK as refactors outside the tested areas stay relatively small since it's still really easy to regress things without realizing. I'd also love to get our test suite to become broad enough that we can heavily refactor some of the nastier components (looking at you `interpreter`)

src/keybindings/keybindings.rs
Valentin271: I agree, I'll add that

Valentin271:
Updated to add logs.

Thanks I'm really happy with how the implementation turned out too.
Understood for the refactors :+1:

PR #226, opened 2024-02-01
Make underlines and strikethroughs respect align

### LINKED ISSUE #141: Underline alignment regression

CosmicHorrorDev:
Resolves #141 

@Valentin271 I got a better idea of the underline code after working on #225. The fix _feels_ hacky, but I'm not really familiar with the renderer and there's a lot of messy code in there to begin with :shrug:

From what I can gather `left` gets set to the x coord of the text area including the `Align`

## Demo

(Includes the use of `--page-width` to check that too)

https://github.com/trimental/inlyne/assets/30302768/107b09f1-e6f4-490e-9484-b259fb3fde88

### DIFF
--- a/src/text.rs
+++ b/src/text.rs
@@ -92,6 +92,7 @@ impl Default for TextBox {
     }
 }
 
+#[derive(Clone)]
 pub struct CachedTextArea {
     key: KeyHash,
     left: f32,
@@ -101,10 +102,6 @@ pub struct CachedTextArea {
 }
 
 impl CachedTextArea {
-    pub fn left(&self) -> f32 {
-        self.left
-    }
-
     pub fn text_area<'a>(&self, cache: &'a TextCache) -> TextArea<'a> {
         TextArea {
             buffer: cache.get(&self.key).expect("Get cached buffer"),
@@ -301,7 +298,7 @@ impl TextBox {
         screen_position: Point,
         bounds: Size,
         zoom: f32,
-        left: f32,
+        text_area: &CachedTextArea,
     ) -> Vec<Line> {
         fn push_line_segment(
             lines: &mut Vec<ThinLine>,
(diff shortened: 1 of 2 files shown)


src/text.rs
```diff
@@ -372,7 +377,7 @@ impl TextBox {
                 let start_cursor = Cursor::new(line.line_i, range.start);
                 let end_cursor = Cursor::new(line.line_i, range.end);
                 if let Some((highlight_x, highlight_w)) = line.highlight(start_cursor, end_cursor) {
-                    let x = screen_position.0 + highlight_x;
+                    let x = left + highlight_x;
```
CosmicHorrorDev, line 380: `left` is already `screen_position.0` for `Align::Left`, but correctly handles the offset for `Center` and `Right` as well (this is why the underlines were stuck left-aligned)

Valentin271 (commented):
> there's a lot of messy code in there to begin with

Yep felt that too.

This does feel a bit hacky to me too. Ideally `render_lines` would take a `CachedTextArea` or similar and not every detail about it.
But still this is way better than what I was trying to do on my fork.

CosmicHorrorDev:
> This does feel a bit hacky to me too. Ideally `render_lines` would take a `CachedTextArea` or similar and not every detail about it.

Good point! Done and done

Valentin271 (approved):
Looks good to me

PR #235, opened 2024-02-11
Make the HTML interpreter more approachable

CosmicHorrorDev:
This is a big bundle of changes that tries to make the code for the interpreter a decent bit more approachable (although it still has a long way to go in terms of larger refactors)

~~Draft since I'm going to go back through the history and clean things up before merging. There were several bugs that got fixed during the changes~~

This contains a myriad of various improvements (more exhaustive matching, logging more interesting things, reworking and adding a whole slew of constructors, etc.) all with the goal of making the `impl TokenSink for HtmlInterpreter` implementation more approachable


CosmicHorrorDev:
@Valentin271 Not sure if you're interested in reviewing, but I would highly recommend switching to hide whitespace when looking through `interpreter/mod.rs` or else the diff is abysmal (not sure if that's a beta feature or not)

image

src/interpreter/html.rs
```diff
@@ -248,3 +252,120 @@ pub enum Element {
...
+    Summary,
+    Table,
+    TableBody,
+    TableDataCell,
+    TableHead,
+    TableHeader,
+    TableRow,
+    Underline,
+    UnorderedList,
+}
```
Valentin271, line 325: This is an important and large part of the interpreter, I'd suggest moving it to another subfile.
`Element` could also benefit from moving to another subfile

Valentin271 (approved):
This is a good first step towards simplifying this.
In general this looks good, my only comment would be about trying to split types in more files, to have more concise files.

CosmicHorrorDev:
Agreed. Split them both into their own files 👍

PR #258, opened 2024-02-26
feat: History and navigation via shortcuts.

kokoISnoTarget:
Created a way to navigate between files opened by clicking on links, as mentioned in #53.
This would also solve #162.
Current shortcuts are `Alt+Left` for the previous file and `Alt+Right` for the next file.

I was not sure about vim like binding, but was thinking of something like `bn` and `bp`.

Please leave any feedback.

### DIFF
--- a/src/keybindings/defaults.rs
+++ b/src/keybindings/defaults.rs
@@ -127,5 +127,21 @@ pub fn defaults() -> Vec<(Action, KeyCombo)> {
                 ModifiedKey(Key::from(VirtKey::Q), ModifiersState::SHIFT),
             ]),
         ),
+        // Navigate to next file: Ctrl+Right
+        (
+            Action::Navigate(Navigation::Next),
+            KeyCombo(vec![
+                ModifiedKey::from(VirtKey::B),
+                ModifiedKey::from(VirtKey::N),
+            ]),
+        ),
+        // Navigate to previous file: Ctrl+Left
+        (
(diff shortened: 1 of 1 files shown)


CosmicHorrorDev:
Thanks for the PR! I'm on vacation right now, but I should be free to review this later this week

src/history.rs
```diff
@@ -0,0 +1,46 @@
...
+        self.history.push(file_path);
+        self.index += 1;
+    }
+    pub fn get_path(&self) -> &PathBuf {
+        self.history
+            .get(self.index)
+            .expect("History should be bound checked for all possible indexes.")
+    }
+
+    pub fn next(&mut self) -> Option<&PathBuf> {
```
CosmicHorrorDev, line 32:
```suggestion
    #[allow(clippy::should_implement_trait)]
    pub fn next(&mut self) -> Option<&PathBuf> {
```

I think the name is fine here

CosmicHorrorDev (approved):
The review comment should get CI passing, but everything else looks great for an initial implementation! Congratz on your first PR here :rocket:

CosmicHorrorDev:
> I was not sure about vim like binding, but was thinking of something like bn and bp.

Buffer next and buffer previous were also my first thought on a vim-keybinding for it. I also don't think that `b` by itself would get used for anything, so that should be totally free in terms of adding a keybind

CosmicHorrorDev:
Looks great! Thanks again

PR #281, opened 2024-03-22
Add caching for images

### LINKED ISSUE #234: Cache downloaded images
We should respect some of the basic properties for image caching (with an easy way to clear the cache), so that we can avoid repeating a bunch of image calls if we have a valid cached image already. This should also speed up what seems to be the slowest visual part of document load times when we have a warm cache

kokoISnoTarget:
This would close #234, but I'm not quite sure if my implementation is any good.
Further, how should the cache be cleared? Delete the directory? Add a Command line option?

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,16 +20,26 @@ Except 9 times out of 10 you need an entire web browser to quickly open a file..
 Introducing **Inlyne**, a GPU powered yet browserless tool to help you quickly
 view markdown files in the blink of an eye 👀.
 
-## Install
+## Installation
 
-To install just use `cargo install inlyne`, almost everything comes pre-bundled (see [Dependencies](#dependencies)).
-Alternatively you can install through various package managers.
+`inlyne` is already present in a handful of different package repos.
+Installation should be mostly painless as _most_ things come pre-bundled.
 
 [![Packaging status](https://repology.org/badge/vertical-allrepos/inlyne.svg?exclude_unsupported=1)](https://repology.org/project/inlyne/versions)
 
-### Dependencies
+### Building from Source
 
-The `fontconfig` package from your distribution needs to be installed. See https://github.com/yeslogic/fontconfig-rs#dependencies.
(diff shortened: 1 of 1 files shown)


CosmicHorrorDev:
Thanks for the pull request! I should have added more info to the original issue since I've slowly been thinking through things, but implementation-wise I was thinking of using an embedded database like [`redb`](https://crates.io/crates/redb) to act as the cache itself (embedded database as in mostly just persistent btrees. Not full SQL like sqlite)

In regard to detecting stale entries:

- for local files this should be as simple as storing and then comparing the modified times
- for URLs this gets a lot more complicated when you start considering [Cache-Control](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control) stuff

I also think it would be good to have some kind of garbage control over entries in the cache. Either a time to live or some overall size limit would be (maybe both?)

And then it would likely be good to have some command line option to clear the cache too since we don't want users to _have_ to mess with deleting files on their own

---

The initial implementation doesn't have to handle all of this of course. That's just my rough vision for the future

CosmicHorrorDev:
Oh and before I forget we should probably also version the data format used for the image data before we want to change the format and run into issues in the future. We currently just LZ4 compress the raw image data, so storing an initial byte for the format before each entry should work. E.g.

```
#[repr(u8)]
enum Format {
    RawLz4Compressed = 0,
}

// Boilerplate `u8` <-> `Format` `From` impls
```

kokoISnoTarget:
> And then it would likely be good to have some command line option to clear the cache too since we don't want users to _have_ to mess with deleting files on their own

Maybe we should rework the CLI first, so something like this can be added easily later? I could open an issue for that.

CosmicHorrorDev:
Sure! That would be a good start. There's all the stuff I mentioned in https://github.com/Inlyne-Project/inlyne/issues/131#issuecomment-2016552309 that I think would be a good direction to move in

Issue #282, opened 2024-03-23
CLI rework


kokoISnoTarget:
There seems to be a need to rework the CLI, as mentioned in #51, #131 and #281 to add options in the future.

I could try to implement that.

CosmicHorrorDev:
Looks like this should be done now :+1:

(I forgot to link the issue)

PR #284, opened 2024-03-23
CLI rewrite

kokoISnoTarget:
For #282.
At this point, none of the (CLI)test work (nor compile).
Screenshot_2024-03-23-22-18-27_5120x1263
But now it should be pretty easy to add new functionality to the CLI.


CosmicHorrorDev:
Yup, but only in the case of how `inlyne view` is handled (before it would open a markdown file named "view", while now it runs the subcommand)

kokoISnoTarget:
Oh, you're right.

CosmicHorrorDev (approved):
Everything looks great! This is more or less exactly how I was thinking about implementing things :)

PR #285, opened 2024-03-24
Add the automatic creation of config files and the ability to open them with the CLI.

kokoISnoTarget:
In #131 the two things (to automatically create a config file and the ability to open them) were mentioned.
This pr adds the CLI option `inlyne config open` and it places the [inlyne.default.toml](https://github.com/Inlyne-Project/inlyne/blob/main/inlyne.default.toml) as `inlyne.toml` in the config directory when no config exists.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -782,6 +782,10 @@ fn main() -> anyhow::Result<()> {
                 .join("inlyne.toml");
 
             if !config_path.is_file() {
+                tracing::info!(
+                    "No config found. Creating a new config at: {}",
+                    config_path.display()
+                );
                 Config::create_default_config(&config_path)?;
             }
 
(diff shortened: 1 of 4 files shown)


src/main.rs
```diff
@@ -775,6 +775,18 @@ fn main() -> anyhow::Result<()> {
...
         }
+        Commands::Config(ConfigCmd::Open) => {
+            let config_path = dirs::config_dir()
+                .context("Failed to find the configuration directory")?
+                .join("inlyne")
+                .join("inlyne.toml");
+
+            if !config_path.is_file() {
+                Config::create_default_config(&config_path)?;
+            }
```
CosmicHorrorDev, line 786: Probably worth an INFO log when it creates a new default config

CosmicHorrorDev (approved):
Looks great as usual!

PR #289, opened 2024-03-26
feat: Setup initial metrics infrastructure

CosmicHorrorDev:
This adds something that I've been wanting implemented for a while. Metrics recording and emission :tada:

There's a new section in the config file that sets the metrics recorder

```toml
[debug]
metrics = "log"
metrics = "tcp"
```

It's unset by default which uses the noop recorder (aka as cheap as it can get)

## Log Recorder

(this one took a bit of work since it implements a recorder from "scratch")

When it's set to `"log"` the metrics start getting recorded as tracing events with some metadata attached on the span. This makes it possible to filter things pretty dynamically, and is accessible as a user doesn't need to download anything else to see the values

image

## TCP Recorder

When it's set to `"tcp"` the metrics get emitted following the default [`metrics-exporter-tcp`](https://docs.rs/metrics-exporter-tcp/0.9.0/metrics_exporter_tcp/) configuration. This means that anything that knows how to consume that can handle displaying the data live. Currently the only main program I know of is the [`metrics-observer`](https://crates.io/crates/metrics-observer) binary crate, but it's unfortunately a bit limited

https://github.com/Inlyne-Project/inlyne/assets/30302768/d1f684e0-30cd-49e1-a0c5-d32dd7d2fe99

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -506,7 +506,7 @@ version = "4.5.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "90239a040c80f5e14809ca132ddc4176ab33d5e17e49691793296e3fcb34d72f"
 dependencies = [
- "heck",
+ "heck 0.5.0",
  "proc-macro2",
  "quote",
  "syn 2.0.52",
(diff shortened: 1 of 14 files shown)


CosmicHorrorDev:
I think it'd be worth investigating different recorders to see if we can find more featureful ones, and document any relevant findings (including the new `cfg` and `config` stuff from this PR)

Issue #173, opened 2023-11-22
Add window display parameters


DeepTulip:
Is it possible to add a parameter that defines the position and size of the window?
For example --win-pos=0,0,2000,1000

It's a bit annoying to have to manually resize the window every time you start inlyne.

--
BTW, I encountered a small reported error:
```
thread '<unnamed>' panicked at src\text.rs:211:17:
attempt to subtract with overflow
```

The fix is:
```
if index > 0
{
index -= 1;
};
```

--

Thank you very much for this software, it is exactly what I was expecting.

PR #290, opened 2024-03-27
feat: Add window position and size configuration options.

kokoISnoTarget:
This would add the possibility to set the window position with `-p` or `--position` and the window size with `-g` or `--size` as requested in #173.
And it would add the ability to do these via the config file:
```toml
[position]
x = 500
y = 200

[size]
width = 600
height = 500
```

### DIFF
--- a/inlyne.default.toml
+++ b/inlyne.default.toml
@@ -20,6 +20,16 @@ theme = "Auto"
 # lines-to-scroll = 4.5
 lines-to-scroll = 3.0
 
+# Position of the window
+# [position]
+# x = 500
+# y = 200
+
+# Geomertry/Size of the window in pixels
+# [size]
+# width = 600
+# height = 500
+
 # The light and dark themes can be customized as well
 # Both the light and dark theme colors can be fully customized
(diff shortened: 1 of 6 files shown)


CosmicHorrorDev (commented):
Implementation-wise everything looks good. I think the only thing I want to bikeshed is the config and CLI specifics. For the config I was thinking more of something like

```toml
[window]
position = [500, 200]
size = [600, 500]
```

For the CLI I think everything looks good. The only thing I'm hesitant on is the `-g` short-flag since `g` doesn't seem like an obvious choice to me. I think if we remove that I'm happy with everything else

kokoISnoTarget:
> For the CLI I think everything looks good. The only thing I'm hesitant on is the -g short-flag since g doesn't seem like an obvious choice to me.

Its original was the short version for geometry and stayed as an artifact of that. But you're right, that's not sensible now.

CosmicHorrorDev (approved):
Makes sense :+1:

Everything looks great now! Thanks as always :D

PR #291, opened 2024-03-30
Add a persistent image cache

### LINKED ISSUE #234: Cache downloaded images

CosmicHorrorDev:
This will eventually resolve #234


CosmicHorrorDev:
Just some random tidbits:

- Useful looking crate for handling HTTP cache semantics ([`http-cache-semantics`](https://crates.io/crates/http-cache-semantics)). Potentially this crate as a higher level one depending on how much `async` is a requirement there ([`http-cache`](https://docs.rs/http-cache/0.19.0/http_cache/))
- There's some interop required to get the right info out of `ureq` but that's not the worst
- The DB format is a bit more static with `redb` than it would be with some SQL DB that can be changed with migrations. I think we can have some consistent header with the internal version and TTL so that we can mix-n-match versions a bit as long as the table layout doesn't change

CosmicHorrorDev:
```text
thread 'image::cache::tests::layers' panicked at src/image/cache/tests.rs:306:66:
called `Result::unwrap()` on an `Err` value: Failed creating database

Caused by:
    Database already open. Cannot acquire lock.


failures:
    image::cache::tests::layers
```

Guess who has two thumbs and gets to port things away from `redb`? :+1::disappointed::+1:

CosmicHorrorDev:
Still a little ways to go, but getting closer at least :tada:

```text
running 1 test
test image::cache::tests::layers ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 52 filtered out; finished in 0.03s
```

CosmicHorrorDev:
The test suite is a good representation of the current state of the PR

```console
$ cargo t image::cache
...

running 17 tests
test image::cache::tests::global_ttl ... ignored, TODO: waiting for garbage collection
test image::cache::tests::lru ... ignored, TODO: waiting for garbage collection
test image::cache::tests::mutli_client_mash ... ignored, TODO: waiting for garbage collection
test image::cache::tests::remote_404_error ... ok
test image::cache::tests::private_cache ... ok
test image::cache::tests::etag_refresh_same ... ok
test image::cache::tests::selectively_stores ... ok
test image::cache::tests::remote_layers ... ok
test image::cache::tests::corrupt_db_entry ... ok
test image::cache::tests::etag_refresh_different ... ok
test image::cache::tests::local_invalidation ... ok
test image::cache::tests::invalid_img ... ok
test image::cache::tests::stats ... ok
test image::cache::tests::local_svg_layers ... ok
test image::cache::tests::local_layers ... ok
test image::cache::tests::past_max_age_refetch ... ok
test image::cache::tests::remote_svg_layers ... ok

test result: ok. 14 passed; 0 failed; 3 ignored; 0 measured; 54 filtered out; finished in 0.27s
```

The only thing left to implement is garbage collection (including probably updating entry's m-times)

## Potential future work

- Integrating _a lot_ more metrics
  - Counters for image loads (from source / l1 cached / l2 cached / errors)
  - Timings for image loads
- Switching DB connections to use a pool to manage re-using connections
- Make the stats output a lot prettier

CosmicHorrorDev:
I think reasonably we could leave the persistent cache disabled by default behind a dev toggle, kick the can on GC stuff and any of the nitty gritty thread pool stuff that we could do, and merge this after replacing the current in-memory cache with the one in this PR

Issue #43, opened 2022-09-11
Support double and triple click for highlighting text


CosmicHorrorDev:
Currently text highlighting only works by dragging a selection. It would be nice to support double and triple click to handle selecting the hovered word and line respectively

PR #295, opened 2024-04-03
Add double and triple click selection

kokoISnoTarget:
This is and draft pr for #43.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -493,7 +493,7 @@ impl Inlyne {
                                         event_loop_proxy
                                             .send_event(InlyneEvent::Reposition)
                                             .unwrap();
-                                        // self.selection.add_position(mouse_position);
+                                        self.selection.add_position(mouse_position);
                                     },
                                     _ => {
                                         self.selection.add_position(mouse_position);
--- a/src/main.rs
+++ b/src/main.rs
@@ -57,7 +57,7 @@ use tracing_subscriber::util::SubscriberInitExt;
 use utils::{ImageCache, Point, Rect, Size};
 
 use crate::opts::{Commands, ConfigCmd, MetricsExporter};
-use crate::selection::{Selection, SelectionKind};
+use crate::selection::Selection;
 use anyhow::Context;
 use clap::Parser;
 use taffy::Taffy;
@@ -425,16 +425,8 @@ impl Inlyne {
                                 * self.renderer.positioner.reserved_height;
                             self.renderer.set_scroll_y(target_scroll);
                             self.window.request_redraw();
-                        } else if let SelectionKind::Drag { start, end } = &mut self.selection.selection {
-                            if mouse_down {
-                                *end = loc;
-                                self.window.request_redraw();
-                            }
-                        } else if let SelectionKind::Start { position, .. } = self.selection.selection {
-                            if mouse_down {
-                                self.selection.selection = SelectionKind::Drag { start: position, end: loc };
-                                self.window.request_redraw();
-                            }
+                        } else if mouse_down && self.selection.handle_drag(loc) {
+                            self.window.request_redraw();
                         }
                         mouse_position = loc;
                     }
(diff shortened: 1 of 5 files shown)


src/main.rs
```diff
@@ -280,9 +284,7 @@ impl Inlyne {
         let mut scrollbar_held = None;
         let mut mouse_down = false;
         let mut modifiers = ModifiersState::empty();
-        let mut last_loc = (0.0, 0.0);
-        let mut selection_cache = String::new();
-        let mut selecting = false;
+        let mut mouse_position: Point = Point::default();
```
CosmicHorrorDev, line 287: Really happy to see better types getting more use :heart:

src/selection.rs
```diff
@@ -0,0 +1,112 @@
+use crate::utils::{dist_between_points, Point};
+use std::time::Instant;
+
+type Milliseconds = u128;
+
+const CLICK_TOLERANCE: Milliseconds = 300;
```
CosmicHorrorDev, line 6:
```suggestion
const CLICK_TOLERANCE: Duration = Duration::from_millis(300);
```

Might as well use a duration here since we can (still requires more changes to anywhere `CLICK_TOLERANCE` gets used though)

src/selection.rs
```diff
@@ -0,0 +1,112 @@
...
+        matches!(self.selection, SelectionKind::None)
+    }
+    pub fn start(&mut self, position: Point) {
+        self.selection = SelectionKind::Start {
+            position,
+            time: Instant::now(),
+        }
+    }
+
+    pub fn handle_drag(&mut self, new_position: Point) -> bool {
```
CosmicHorrorDev, line 54:
```suggestion
    #[must_use]
    pub fn handle_drag(&mut self, new_position: Point) -> bool {
```

Seems easy to forget that this has a meaningful return type

src/text.rs
```diff
@@ -434,65 +427,108 @@ impl TextBox {
...
+                    position.1 - screen_position.1,
+                )?;
+
+                let line = buffer.lines.get(cursor.line)?;
+
+                match mode {
+                    SelectionMode::Word => {
+                        let text = line.text();
+                        if text
+                            .get(cursor.index..cursor.index)?
```
CosmicHorrorDev, line 457: I don't think this does anything since the `.get(cursor.index..cursor.index)` will always be an empty slice. I _think_ this check is to make sure that you can't do a word double-click selection when you're hovering over whitespace, but it still seems to try with the current impl where it winds up selecting words on either side minus the first character on the left/right side's word depending on the cursor's affinity (aka if it's on the left or right side of the whitespace character)

image

src/text.rs
```diff
@@ -434,65 +427,108 @@ impl TextBox {
...
+                        let start = Cursor::new(cursor.line, start_index);
+                        let end = Cursor::new(cursor.line, end_index);
+
+                        (start, end, position.1, position.1)
+                    } // TODO implement
+                    SelectionMode::Line => {
+                        let start = Cursor::new(cursor.line, 0);
+                        let end = Cursor::new(cursor.line, line.text().len());
+                        (start, end, position.1, position.1)
+                    } // TODO implement
```
CosmicHorrorDev, line 482: Left-over comments, or is there still more left to handle?

CosmicHorrorDev (changes requested):
Another issue I noticed, but I'm not sure where in the code is the problem, but it seems like the selections pull in another word if you're double-clicking on a 1-2 character word, but 3+ seems fine

---

Overall most things seem to work great! (like really really great :rocket:)

I left feedback on some bits I noticed and then there's also merge conflicts, but I'm super excited to get this merged

src/text.rs
kokoISnoTarget: Oops

CosmicHorrorDev (approved):
Looks great! Thanks as always :tada:

PR #299, opened 2024-04-06
v0.4.3 backportapalooza

CosmicHorrorDev:
Figured that the next breaking release is far enough away that it's worth backporting some things in the meantime

Planning on cutting the actual release tomorrow after I have time to double check things


CosmicHorrorDev:
Woops. Branch name is wrong

PR #308, opened 2024-04-12
feat: add debug boxes

kokoISnoTarget:
This adds bounding boxes to elements when the `debug` flag is used, as mentioned in https://github.com/Inlyne-Project/inlyne/issues/294#issuecomment-2050745847
Screenshot_2024-04-12-12-43-38_5120x1263
Screenshot_2024-04-12-12-44-23_5120x1263


CosmicHorrorDev (approved):
Looks great, thanks! I'm planning on moving the functionality behind a config value since it's hard to have non-user facing features, so I tend to avoid those when I can

This is the hard part of the changes though, so very happy to have this done :)

PR #314, opened 2024-04-13
feat: scroll when clicking the scrollbar

### LINKED ISSUE #311: Scrollbar only reacts on movement
Currently the scrollbar only reacts when the mouse-cursor is moved; however, this is clunky when it comes to clicking the scrollbar area to jump to a position since it requires shifting the mouse while pressing the button down instead of being able to just click (this is similar to issues we've had in the past, and was the original motivation to storing the mouse cursor's last known position)

https://github.com/Inlyne-Project/inlyne/assets/30302768/f6a6f49d-8cef-4bfa-be0f-de14af6ed458

kokoISnoTarget:
Fixes #311

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -387,9 +387,7 @@ impl Inlyne {
                             .contains(position.into())
                                 && mouse_down)
                         {
-                            let scrollbar_height = (screen_size.1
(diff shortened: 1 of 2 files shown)


CosmicHorrorDev (approved):
Looks good. Thanks!

PR #320, opened 2024-04-17
refactor(tests): Drop wiremock

CosmicHorrorDev:
We really didn't need a lot of what `wiremock` provided since we really only want a small file server, so this PR switches to precisely that


CosmicHorrorDev:
This drops the total number of dependencies in our tree from 442 -> 423

PR #329, opened 2024-05-12
Windows CI is hanging

CosmicHorrorDev:
Looks like windows CI has gotten in the habit of hanging with the test suite. Time to see if it's a new problem with us or a new problem with GH

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -53,8 +53,8 @@ jobs:
       - name: Cache
         uses: Swatinem/rust-cache@v2
 
-      - name: Linting
-        run: cargo clippy --workspace --no-default-features --features x11 -- --deny warnings
+      # - name: Linting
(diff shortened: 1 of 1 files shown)


CosmicHorrorDev:
Whelp that's the same kind of failure we've been seeing lately all the way back on e26ee7da4d7a0adfdc00991351269f240ea66651, so it looks like something outside of our domain changed to cause this to start hanging. Wasn't able to reproduce the hang when running the test suite locally on windows 10 either. Guess I'll just ignore it for now and hope that it eventually goes away

CosmicHorrorDev:
It does look like the hangs are much less frequent back in this point in history though

The biggest change to the test suite since then was the switch from `wiremock` to `tiny_http` for setting up the fake servers for us to interact with. `wiremock` tries to reuse a pool of servers while we spin up a fresh server per test when using `tiny_http`. My hunch is that there's some implicit cleanup happening that can hang and spinning up more servers is increasing that chance

PR #331, opened 2024-05-19
refactor: Re-organize various test utilities

CosmicHorrorDev:


CosmicHorrorDev:
Looks like the hang is due to the various image servers, so looks like it's time to try refactoring to share a single one

CosmicHorrorDev:
Refactoring things to share a single server for all tests instead of spinning up isolated servers per test appears to have fixed the hang with windows CI :face_exhaling:

PR #339, opened 2024-06-06
fix: selection bugs

### LINKED ISSUE #330: Click selection broken for long texts
Double-clicking text which is at least one window height away from its first row is broken.
Snapshot_2024-05-15_19-04-08
Snapshot_2024-05-15_19-08-33
(Highlighted part is not selectable via clicking)

kokoISnoTarget:
Fixes #330 and another bug (look commit msg)


CosmicHorrorDev (approved):
Looks good, thanks!

Issue #356, opened 2024-08-29
possible rust 1.80 build issues


dezren39:
hello i am unable to build this pkg from nixpkgs master branch right now.

i believe it is due to https://github.com/NixOS/nixpkgs/issues/332957

i attempted to confirm this myself, but had a couple issues working through building in general, with cargo wanting a 'wayland-client' to be available. this is likely just because i don't have my enviroment properly setup for building inlyne.

in other packages they suggest updating 'time' but i don't think that is the issue here.

CosmicHorrorDev:
It's possible that there could be issues as the last release looks like it's on an affected time version

```console
❯ cargo tree -p time
time v0.3.29
├── deranged v0.3.8
├── itoa v1.0.9
└── time-core v0.1.2
```

which predates the PR to `time` fixing it (0.3.29 was released 7 months ago and the PR was merged 5 months ago), and I can confirm that I'm getting the expected error when trying to build the last `inlyne` release with a more recent toolchain

```text
   Compiling enumflags2 v0.7.7
   Compiling gpu-descriptor v0.2.3
error[E0282]: type annotations needed for `Box<_>`
  --> /home/wintermute/.cargo/registry/src/index.crates.io-6f17d22bba15001f/time-0.3.29/src/format_description/parse/mod.rs:83:9
   |
83 |     let items = format_items
   |         ^^^^^
...
86 |     Ok(items.into())
   |              ---- type must be known at this point
   |
help: consider giving `items` an explicit type, where the placeholders `_` are specified
   |
83 |     let items: Box<_> = format_items
   |              ++++++++

   Compiling zvariant_derive v3.14.0
   Compiling darling v0.14.4
```

The `main` branch here is already updated to the latest time, although I won't be able to get a release out for likely a couple more weeks at least

As per the `wayland-client` issues that's an expected dep if the `wayland` feature is enabled which it is by default. There's are details on what dependencies are required in the [_Building from Souce_ section of the README](https://github.com/Inlyne-Project/inlyne?tab=readme-ov-file#building-from-source)

CosmicHorrorDev:
Eh on second thought we already have a branch for 0.4, so I can push out a patch release with an updated `time` version later today

CosmicHorrorDev:
A new release with an updated `time` version is now out which @r-ryantm will likely pick up and open a PR to nixpkgs soon to update to the latest version

CosmicHorrorDev:
[...]

PR #357, opened 2024-08-29
deps: `$ cargo update -p time`

CosmicHorrorDev:
Related #356


CosmicHorrorDev:
Adding a backlink so that were included in the fallout

rust-lang/rust#125319

PR #361, opened 2024-09-14
Fix: Updating window title if new file loaded

BoolPurist:
Fixes the following [issue](https://github.com/Inlyne-Project/inlyne/issues/360)

Refactors: Encapsulated updating file watcher and window title into one function

Note: This was a regression bug.
In previous versions, this bug did not exit
See the following commit "a6bb495" with version 0.4.1

## Reason why PR on branch v0.4

The current version on crates.io seems to use commit 588687e8eccb2f01573dc5eed7597b62d7253ae4 on this branch.
Therefore, the addressed bug comes up when installing inlyne via the following command 

```txt
cargo install inlyne
```

It also seems that the most recent changes are done on this branch.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -175,16 +175,21 @@ fn root_filepath_to_vcs_dir(path: &Path) -> Option<PathBuf> {
     }
 }
 
+/// Sets window tile to the path of currently loaded file.
+fn set_window_title_with_new_path(window: &Window, new_path: &Path) {
+    match root_filepath_to_vcs_dir(new_path) {
+        Some(path) => window.set_title(&format!("Inlyne - {}", path.to_string_lossy())),
+        None => window.set_title("Inlyne"),
+    }
+}
+
 impl Inlyne {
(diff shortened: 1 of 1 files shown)


CosmicHorrorDev:
Would you mind switching the base branch to `main`? v0.4 only exists for backporting fixes and won't be used after the pending 0.5.0 release

CosmicHorrorDev:
Well that button didn't do what I wanted. I'll work on fixing the conflicts

CosmicHorrorDev:
Ahhhh. I somehow forgot that this was already working as intended on main with

- #274

so this would strictly be for the `v0.4` branch :sweat_smile:. I could merge this as a backport, but I'm not sure if it's worth it or not since it'll also be fixed once 0.5.0 is released

BoolPurist:
I think this path is then obsolete in that case. No biggie. Say the word and I close this PR.

CosmicHorrorDev:
Fair enough. I'm sorry that I didn't realize earlier, thanks for your contribution nonetheless

PR #395, opened 2025-03-05
Make window decorations optional and use a better CSD on Wayland

Koranir:
This PR lets you pass the `-d` or `--decorations` boolean argument to enable/disable decorations, and enables winit's sctk-adwaita CSDs for Wayland compositors that don't use SSDs.

| prev (ugly) | this PR (CSD) | no decorations |
|-------------|---------------|----------------|
| image | image | image |

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,22 @@
 version = 3
 
 [[package]]
+name = "ab_glyph"
+version = "0.2.29"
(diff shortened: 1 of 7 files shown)


CosmicHorrorDev (approved):
Looks great. Thanks and congrats on your first PR! 🚀

PR #413, opened 2025-08-01
Solve the issue with long words going offscreen

luizvbo:
Solves https://github.com/Inlyne-Project/inlyne/issues/351

### DIFF
--- a/src/text.rs
+++ b/src/text.rs
@@ -7,8 +7,8 @@ use std::sync::Arc;
 
 use fxhash::{FxHashMap, FxHashSet};
 use glyphon::{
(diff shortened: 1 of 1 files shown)


CosmicHorrorDev:
It looks like we want `Wrap::WordOrGlyph` that first tries to wrap based on words, and if that fails to fit on one line falls back to wrapping by glyph instead: _however_, that was added in `cosmic-text v0.11.0` and we're currently stuck on `v0.9.0` until we handle updating those dependencies

luizvbo:
I will look into that this week. Thansk

PR #419, opened 2025-10-06
Handle `wgpu::SurfaceError` gracefully during rendering.

glls:
Check if renderer throw an error to prevent abnormal app termination.

Fixes: https://github.com/Inlyne-Project/inlyne/issues/418

### DIFF
--- a/src/renderer.rs
+++ b/src/renderer.rs
@@ -698,7 +698,9 @@ impl Renderer {
             }
             std::result::Result::Err(wgpu::SurfaceError::OutOfMemory) => {
                 // Fatal; bubble up to allow graceful shutdown.
-                Err(anyhow::anyhow!("Out of memory while acquiring swap chain texture"))?
+                Err(anyhow::anyhow!(
+                    "Out of memory while acquiring swap chain texture"
+                ))?
             }
         };
         let view = frame


CosmicHorrorDev:
Looks like the typo failure is from a `typos` update where it's picking up an existing typo. I can push a fix for that sometime over the next couple of days

Thanks for the PR!

src/renderer.rs
```diff
@@ -683,10 +683,24 @@ impl Renderer {
...
+                // Surface was lost or outdated (e.g., window resized or moved to a different output)
+                // Reconfigure and skip this frame; a new frame will be requested automatically.
+                self.surface.configure(&self.device, &self.config);
+                return Ok(());
+            }
+            std::result::Result::Err(wgpu::SurfaceError::Timeout) => {
+                // GPU is busy; skip this frame.
+                return Ok(());
+            }
+            std::result::Result::Err(wgpu::SurfaceError::OutOfMemory) => {
```
CosmicHorrorDev, line 699:
```suggestion
            Ok(frame) => frame,
            Err(wgpu::SurfaceError::Lost | wgpu::SurgaceError::Outdated) => {
                // Surface was lost or outdated (e.g., window resized or moved to a different output)
                // Reconfigure and skip this frame; a new frame will be requested automatically.
                self.surface.configure(&self.device, &self.config);
                return Ok(());
            }
            Err(wgpu::SurfaceError::Timeout) => {
                // GPU is busy; skip this frame.
                return Ok(());
            }
            Err(wgpu::SurfaceError::OutOfMemory) => {
```

Just two small nits

- You can match directly on `Ok` and `Err` instead of specifying the full path
- `Err(wgpu::SurfaceError::Lost) | Err(wgpu::SurgaceError::Outdated)` is the same as `Err(wgpu::SurfaceError::Lost | wgpu::SurgaceError::Outdated)`

CosmicHorrorDev (commented):
Just some small style stuff, but otherwise things look great! The typo that's causing CI to fail is fixed in `main` if you want to update on top of that

PR #459, opened 2026-06-12
deps: update taffy to 0.11.0

### LINKED ISSUE #438: Specific table size/spacing causes freeze
Seemingly random table entries cause inlyne to freeze upon launch. Either inlyne can render everything up to the table but not including it, can only create the window, or inlyne simply doesn't create a window at all. In all of these cases it requires a Ctrl+C to stop. It also uses 100% of a CPU core until killed

I've attached 3 files that cause a freeze, and 3 files that don't cause a freeze. All of the files with very similar content.

I've also attached a backtrace obtained via rust-gdb during the freeze

My system uses the JetBrainsMono Nerd Font as a default system font. In the case that this is indeed a specific size/spacing issue.

[freeze.md](https://github.com/user-attachments/files/27172898/freeze.md)
[freeze2.md](https://github.com/user-attachments/files/27172900/freeze2.md)
[freeze3.md](https://github.com/user-attachments/files/27172896/freeze3.md)

[works.md](https://github.com/user-attachments/files/27172899/works.md)
[works2.md](https://github.com/user-attachments/files/27172894/works2.md)
[works3.md](https://github.com/user-attachments/files/27172895/works3.md)

[gdb.txt](https://github.com/user-attachments/files/27172897/gdb.txt)

kokoISnoTarget:
This should fix #438

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1710,9 +1710,9 @@ dependencies = [
 
 [[package]]
 name = "grid"
-version = "0.10.0"
+version = "1.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "eec1c01eb1de97451ee0d60de7d81cf1e72aabefb021616027f3d1c3ec1c723c"
+checksum = "b40ca9252762c466af32d0b1002e91e4e1bc5398f77455e55474deb466355ff5"
 
 [[package]]
 name = "half"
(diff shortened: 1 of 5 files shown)


CosmicHorrorDev:
glad to see you're still around @kokoISnoTarget! i'll review when i get time (working lots of long hours lately 😪)

CosmicHorrorDev (approved):
looks great! thanks 🎉 

i wasn't able to reproduce the issue while resizing on either `main` or this PR, but i believe it should be fixed. i have run into it on another setup in the past

PR #462, opened 2026-06-24
deps: downgrade copypasta to 0.8.2 to fix wayland-client version conf

khoroo:
When built with the `wayland` feature, `copypasta 0.10.x` pulls in `smithay-clipboard 0.7.x`, which requires `wayland-client 0.31` / `smithay-client-toolkit 0.20`. This conflicts with winit `wayland-client 0.29` / `sctk 0.16`, causing a runtime crash:

  [wayland-client error] Attempted to dispatch unknown opcode 0 for wl_shm

Downgrading to `copypasta 0.8.2` pins `smithay-clipboard to 0.6.x`, which shares the  `wayland-client 0.29` / `sctk 0.16` versions as winit.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -187,7 +187,7 @@ checksum = "0ae92a5119aa49cdbcf6b9f893fe4e1d98b04ccbf82ee0584ad948a44a734dea"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.117",
+ "syn 2.0.118",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


CosmicHorrorDev:
grrrrr runtime crash. i'll try to take a look tomorrow

CosmicHorrorDev (approved):
i wasn't able to reproduce the craah locally, but your reasoning sounds sound enough based on the diff of `Cargo.lock`

PR #465, opened 2026-07-14
Correctly track relative links when navigating

### LINKED ISSUE #398: Navigation fails with relative paths
It seems all the navigations take the original document path as reference. So let's say you open README.md that references doc/SUMMARY.md , that will work fine in inlyne, however, if you click a link to intro/introduction.md on the SUMMARY.md it will try to navigate to intro from where README.md is, so intro/introduction.md, instead of navigating from where SUMMARY.md is, doc/intro/introduction.md.

durbanlegend:
`Navigation fails with relative paths`

Ensure that `std::env::set_current_dir` is called when following a relative .md link. Canonicalize the path first to ensure it's unambiguously correct for the file watcher and the make_next history function.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -480,8 +480,19 @@ impl Inlyne {
                                             } else {
                                                 match read_to_string(&path) {
                                                     Ok(contents) => {
-                                                        self.update_file(&path, contents);
-                                                        self.opts.history.make_next(path);
+                                                        // Canonicalize before changing the CWD so
+                                                        // that the watcher and history always store
+                                                        // absolute paths, independent of CWD.
+                                                        let abs_path = path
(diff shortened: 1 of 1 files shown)


CosmicHorrorDev:
will take a look tomorrow, thanks!

CosmicHorrorDev:
pulled the second commit into its own PR (#467). will take a closer look at the other bits tomorrow :fairy:

CosmicHorrorDev (approved):
tweaked the commit message, but all the code looks great. thanks!

tested with the reproduction listed in the issues. `main` fails and this PR works perfectly including the file watcher

durbanlegend:
Thank you!

Issue #471, opened 2026-08-02
`inlyne config open` fails when the config directory doesn't exist


mmclinton:
I ran into this while working on #354. It's independent of that rename, so I'm splitting it out.

On a machine that has never run `inlyne view`, `inlyne config open` opens the editor with the default config, then fails to save it:

```console
$ XDG_CONFIG_HOME=$(mktemp -d) inlyne config open
# editor opens on the default config, write and quit
Error: No such file or directory (os error 2)
```

Reproduced on `main` (e70ddb1) on Linux and MacOS. v0.5.0 through v0.5.2 are affected.

The `Commands::Config(ConfigCmd::Open)` arm ends with `src/main.rs:827`:

```rust
std::fs::write(config_path, new_config)?;
```

`std::fs::write` won't create `<config dir>/inlyne/`, so on a fresh install the write fails with a bare `io::Error` naming neither the path it tried nor the missing directory. Startup doesn't hit this because `Config::load_from_system` goes through `create_default_config`, which does `create_dir_all` on the parent first.

The edit isn't truly lost. `.disable_cleanup(true)` on the `edit::Builder` leaves the temp file behind, but its path is never printed, so there's no way for the user to know that.

Either fix works:

- `create_dir_all` on the parent before the write, or
- reuse `Config::create_default_config` for the case where the file doesn't exist yet

and add `.context()` to the write either way so the error names the path.

Happy to open a PR if you'd like.

PR #472, opened 2026-08-02
refactor: prefer `config.toml` over `inlyne.toml`

### LINKED ISSUE #354: Rename the config file `inlyne.toml` -> `config.toml`
It seems a bit much to have the config file use the program's name when it's already in a directory called `inlyne/`. Might as well rename it to the more obvious `config.toml` (although we should still support `inlyne.toml` as well for back compat)

mmclinton:
Closes #354

New installs get `config.toml`. An existing `inlyne.toml` keeps working indefinitely. It's read and written in place, so there's nothing to rename. When both are present `config.toml` wins and the ignored one gets named in a warning.

Resolution is shared between startup and `config open` rather than duplicated, since `config open` writes back to the path it resolves. Had only the load path learned the new name, `config open` would have edited one file while startup read the other.

Unit tests cover the four resolution cases (neither file present, legacy only, current only, both) and that a *directory* named `config.toml` doesn't shadow a real `inlyne.toml`. Also checked by hand on Linux and macOS that either name is picked up on its own, and that `config.toml` wins when both are there:

```console
2026-08-02T20:38:59.210662Z  WARN inlyne::opts::config: Ignoring '/Users/mmc/Library/Application Support/inlyne/inlyne.toml' in favor of '/Users/mmc/Library/Application Support/inlyne/config.toml'
```

I ran into #471 while working on this. It's a pre-existing bug in the same config open arm and independent of the rename, so it's not included here.


mmclinton:
Heads up: I've opened #473 (fixes #471, `inlyne config open` failing on a fresh install). It touches the same two spots this PR does, in `src/opts/config.rs` and the `config open` arm of `main.rs`, so GitHub will flag whichever of the two lands second.

They're independent. No ordering requirement, merge them in whatever order suits you. I'll rebase the second one; what that rebase involves is spelled out in #473's description.

